PR #16, opened 2017-11-01
Fix memory mapped shows old content: Issue 14

gamlerhart:
- Flushes / Invalidates the page cache of the written section
- Enough to fix https://github.com/lundman/ZFSin/issues/14

### DIFF
--- a/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
+++ b/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
@@ -2135,6 +2135,19 @@ NTSTATUS fs_write(PDEVICE_OBJECT DeviceObject, PIRP Irp, PIO_STACK_LOCATION IrpS
 
 	uio_free(uio);
 
+
+	if(!Status) {
+	    IO_STATUS_BLOCK IoStatus = { 0 };
+		// For memory mapped files: flush out page cache of written section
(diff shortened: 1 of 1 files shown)


lundman:
Thanks for looking into this. Are we considering this a temporary solution until cache update can be investigated, or a final one? What does fastfat do?

lundman:
Looks like fastfat can sometimes call `CcCoherencyFlushAndPurgeCache()` so that could be a valid work around. The proper path seems to call either `CcCopyWriteEx()` or `CcPrepareMdlWrite()` for cached writes. Which we should probably investigate in the future.


While there, looks like we should tell cache about file size changes by using:

```
  if (CcIsFileCached(FileObject)) {
     CcSetFileSizes( FileObject, (PCC_FILE_SIZES)&FcbOrDcb->Header.AllocationSize );
```

.. or rather, update the ZFS macros `vn_has_cached_data` to use `CcIsFileCached` and `vnode_pager_setsize` to `CcSetFileSizes`.

gamlerhart:
Thanks for the hints....I'll look into it when I've time.  (All new to me, Windows-Kernel / zfs code base ^0^)

lundman:
I think I'll go ahead and approve of this PR, so it at least works for now, and create an Issue to make the code update cache in write.

Issue #19, opened 2017-11-07
Deleting files does not reclaim space in Explorer


gamlerhart:
Reproduce:
    - Create a bunch of files
    - Delete them
    - Files are gone, but space isn't freed up.

More investigation will follow.

lundman:
How are you measuring the space? zfs list and zpool list should eventually show the disk reclaimed, albeit a bit delayed, since deletion is async.

If you are looking in Explorer etc, I'm sure that won't update as there is no code for it..

gamlerhart:
Clarification: Does not reclaim space in Explorer. (Hmm, unless it's very delayed)
Didn't check zpool/zfs.

Will check again / investigate more.

gamlerhart:
Also doesn't seem to reclaim space in zfs list / zpool.

From logs:
`IRP_MJ_CLEANUP: iocount 0 usecount 2
+zget 4
delete_entry: deleting '\test-file.data'
+zget 60
vnop_remove: may_delete_now is 0, delete_now 0
delete_entry: returning 0`

More investigation will follow. Looks like vnode is consider in use / or cached. (https://github.com/lundman/ZFSin/blob/master/ZFSin/zfs/module/zfs/zfs_vnops.c#L1971)

lundman:
Not sure about that '\\' in the filename, but it is probably just cosmetics.

The real issue is, if `delete_now == 0` it will skip immediate deletion, and place the item on the `unlinked_drain` list, which is purged on next **import**. That is normal for XATTRs, but regular files, we want `delete_now` to be set, so it enters the code that calls `vnode_recycle()` and `zfs_znode_delete()`. 

That is based on `may_delete_now` and the status of `vnode_isinuse()` which probably means it is not 0. But I think that tracks, since Windows opens the file with `delete on close`, so the vnode_isinuse() check probably needs to check iocount is 1 instead of 0.

gamlerhart:
Fixed on master

PR #21, opened 2017-11-11
Potential fix issue #19 Deleting files does not reclaim space in E

gamlerhart:
Potential fix for https://github.com/openzfsonwindows/ZFSin/issues/19. 

Feels icky and hacky.
- In the Windows cleanup routine we may `vnode_recycle` & `zfs_znode_delete`. Then the vnode is marked as `DEAD`.
- On the way out and on close, we now check if the `vnode` is dead.
- Kind of strange to have a 'dead' vnode.
- Also the vnode_isinuse checks seem 'icky'.

Probably more correct solution?
- Mark the file as pending delete. Maybe on the vnode?
- Then do the `vnode_recycle` & `zfs_znode_delete` on the close, if the inuse count drops to zero.


ZFSin/zfs/module/zfs/zfs_vnops.c
```diff
@@ -2074,7 +2075,6 @@ zfs_remove(vnode_t *dvp, char *name, cred_t *cr, caller_context_t *ct,
 		(void) sa_lookup(zp->z_sa_hdl, SA_ZPL_XATTR(zfsvfs),
 		    &xattr_obj_unlinked, sizeof (xattr_obj_unlinked));
 		delete_now = may_delete_now && !toobig &&
-		    !vnode_isinuse(vp,0) && !vn_has_cached_data(vp) &&
```
gamlerhart: This checks the same condition again as the `may_delete_now` flag. Not sure why? 
I didn't see something which may change this?

ZFSin/zfs/module/zfs/zfs_vnops.c
```diff
@@ -2116,18 +2116,20 @@ zfs_remove(vnode_t *dvp, char *name, cred_t *cr, caller_context_t *ct,
...
 		/*
 		 * Call recycle which will call vnop_reclaim directly if it can
 		 * so tell reclaim to not do anything with this node, so we can
 		 * release it directly. If recycle/reclaim didn't work out, defer
 		 * it by placing it on the unlinked list.
 		 */
-
 		zp->z_fastpath = B_TRUE;
-		if (vnode_recycle(vp) == 1) {
+		if (vnode_recycle(vp) == 0) {
```
gamlerhart: vnode_recycle always return 0 at the moment. On success we zfs_znode_delete it. Right?

lundman:
Yes, the way I suggested initially is not quite right. We should only ever call recycle once, and if a vnode is marked DEAD, we really should not do anything with that.  At the moment, since the deletion requests happens on Open in Windows (which is different to Unix), I made the vnode_setdelete() be a tag for delayed deletion.

But I think maybe the place we currently act on the deleted flag is perhaps not in the correct place. Perhaps where we normally call vnode_recycle, it should check for deleted and call zfs_remove() as needed. It should be more of a question of finding the right place/time to call zfs_remove.

gamlerhart:
Will try moving `vnode_recycle` next to `vnode_recycle`. I hope then it flows more natural with the inuse count and vnode cleanp.

gamlerhart:
Improved this PR. Deleting now on IRP_MJ_CLOSE. Way better:

- The ref-count works out.
- No check needed for DEAD vnodes
- And deletes the znode on delete when possible.

lundman:
Yes, that does look nicer.

We do somewhat end up calling vnode_recycle twice, except that the first call clears out the z_vnode, so the second call in CLEANUP skips the call. 

This is OK for the time being, as it happens, ZFS should not call vnode_recycle in CLEANUP anyway.  Normally the VFS layer will keep the vnodes around even when iocount/usecount is 0,0. And only once the number of vnodes reach maxvnodes, will the VFS layer call vnode_recycle on "elected" free nodes. 

But since the VFS code in this ZFS port is "mostly empty", we manually call vnode_recycle in CLEANUP, which is less efficient, but much simpler. Eventually we'll change this, to enjoy the benefit from cached vnodes.

gamlerhart:
Looks to me vnode_recycle() does not vnode_recycle the vnode at them moment. `vp->v_iocount` isn't `1`. so `vnode_recycle` skips the `kmem_free`.

So, vnode better be cleaned up CLOSE for now. 

Will look into it more.

lundman:
That's odd, the IRPs are OPEN, CLOSE then CLEANUP though right? Or is CLOSE last?

gamlerhart:
In my head it's CREATE, (operations), CLEANUP, CLOSE. I'm not surprised if I'm wrong...zero experience.

https://msdn.microsoft.com/en-us/library/windows/hardware/ff548621(v=vs.85).aspx
> This request normally follows a cleanup request. However, this does not necessarily mean that the close request will be received immediately after the cleanup request.

lundman:
```
dispatcher: enter: major 0: minor 0: IRP_MJ_CREATE
dispatcher: enter: major 18: minor 0: IRP_MJ_CLEANUP
dispatcher: enter: major 2: minor 0: IRP_MJ_CLOSE
```

Well, there you go. I guess I assumed it was like unix, with open, close, then inactive. But Windows appears to be open, cleanup then close. Makes me wonder what the cleanup is actually for.

So yes, recycle should be in CLOSE. If it hasn't been called all this time, it may crash, so if that is the case, leave it for me to figure out....

gamlerhart:
On this PR: 
Improved: Now free vnode on delete in 'happy path'.
Afaik does not properly free vnode when enqueuing into the unlink queue.

lundman:
Code looks great, I'll merge that in a bit.

On the unlinked-drain path - let's see. Unix will call zfs_remove() on the item, then call vnop_recycle to ditch the vnode once the iocount goes 0. So in future once SPL VFS is smarter, we can do that, for now though.. We could put in a check in vnode_put() that if it has gone to 0, and was already deleted, recycle. But I think we can leave that for now. That case should only be for XATTRs of which we support none, yet.

PR #23, opened 2017-11-22
Crude Set FileAllocationInformation implementation

gamlerhart:
- I'm sure what the FS system actually is required to do exactly on this call.
   - Didn't find a clear documentation piece for it.
- Seems that it's a pre-allocation request, to 'reserve' blocks for that file.

Doing the 'same' as `set_end_of_file_information` is 'good enough' to fix explorer.
And to make Win32 `SetEndOfFile` work (at last as called in .NET apps).

### DIFF
--- a/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
+++ b/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
@@ -1899,8 +1899,17 @@ NTSTATUS set_information(PDEVICE_OBJECT DeviceObject, PIRP Irp, PIO_STACK_LOCATI
 	NTSTATUS Status = STATUS_NOT_IMPLEMENTED;
 
 	switch (IrpSp->Parameters.SetFile.FileInformationClass) {
-	case FileAllocationInformation: // set allocation size, refreserve?
-		dprintf("* FileAllocationInformation\n");
+	case FileAllocationInformation: 
+		if (IrpSp->FileObject && IrpSp->FileObject->FsContext) {
+			FILE_ALLOCATION_INFORMATION *feofi = Irp->AssociatedIrp.SystemBuffer;
+			dprintf("* FileAllocationInformation %u\n", feofi->AllocationSize.QuadPart);
+			// This is a noop at the moment. It makes Windows Explorer and apps not crash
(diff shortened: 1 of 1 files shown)


gamlerhart:
From reading (https://msdn.microsoft.com/en-us/library/windows/hardware/ff540232(v=vs.85).aspx), FileAllocationInformation  does not increase the file size.
However, setting a smaller FileAllocationInformation than the file size, will truncate the file.

lundman:
Oops, missed this PR. Ok, so you removed the freesp call from FileEndOfFileInformation, and moved it into SetInformation? Should it not still be in FileEndOfFileInformation? If it isn't meant to extend the size (just truncate) we should perhaps check if request is `< zp->z_size`.

gamlerhart:
The FileAllocationInformation and FileEndOfFileInformation both do the `zfs_freesp`, increasing or truncating the file.

This is `enough' to let Windows Explorer and my tiny test app not fail.

> If it isn't meant to extend the size (just truncate) we should perhaps check if request is < zp->z_size 

Yes...
Sorry, I meant to follow up with more investigations, but didn't get to this week.

gamlerhart:
Changed the 'crude' approach to a no-op operation. It's good makes Windows Explorer and apps not crash.

- The documentation says that the file is truncated when setting a smaller allocation size than file length.
However, NTFS doesn't do that. It keeps the file length etc the same. 
- Overall there doesn't seem a observable effect to the application when using this call. 
   - Except the 'space' reservation / maybe performance missing.

For more correct implementation:
   - Basically need to at least reserve space. Until the file handle is closed.
   - Is there such facility to 'temporary' reserve space?
   - Otherwise: We can at least check if the current reserved space for the file system can hold the reservation.

lundman:
Yes that is fine, doing nothing is quite acceptable, we do the same for `F_RDADVISE` under Unix. 

There is technically a way to refreserve space (as used by the property) but I don't see anything in documentation to `FileAllocationInfomation` that we are to guarantee the space to be available.

gamlerhart:
Space guarantee: Yes, the documentation doesn't say anything, yes.
That's from observing NTFS. It there not enough space for the allocation, an error is returned. 
Otherwise not.

PR #32, opened 2018-05-29
Fix spl_free_thread() when segkmem > 75

maksqwe:
```
if ((segkmem_total_mem_allocated * 100LL / real_total_memory) > 70) {
	new_spl_free -= total_mem_used / 64;

// below condition will never be checked if
// (segkmem_total_mem_allocated * 100LL / real_total_memory)  > 75
} else if ((segkmem_total_mem_allocated * 100LL / real_total_memory) > 75) {
	new_spl_free -= total_mem_used / 32;
	lowmem = TRUE;
}
```

### DIFF
--- a/ZFSin/spl/module/spl/spl-kmem.c
+++ b/ZFSin/spl/module/spl/spl-kmem.c
@@ -4766,9 +4766,10 @@ spl_free_thread()
 
 			// memory footprint has gotten really big, decrease spl_free substantially
 			int64_t total_mem_used = (int64_t) segkmem_total_mem_allocated;
-			if ((segkmem_total_mem_allocated * 100LL / real_total_memory) > 70) {
+			int64_t mem_used_perc = (int64_t)(segkmem_total_mem_allocated * 100LL / real_total_memory);
+			if (mem_used_perc > 70 && mem_used_perc <= 75) {
 				new_spl_free -= total_mem_used / 64;
-			} else if ((segkmem_total_mem_allocated * 100LL / real_total_memory) > 75) {
+			} else if (mem_used_perc > 75) {
 				new_spl_free -= total_mem_used / 32;
 				lowmem = TRUE;
 			}


lundman:
Hey @rottegift I think this would also be true in osx: https://github.com/openzfsonosx/spl/blob/master/module/spl/spl-kmem.c#L4688  right?

rottegift:
It's not *exactly* a bug, because segkmem_total_mem_allocated can be (atomically) modified in another thread, leading to a change between the two tests.   Admittedly this is an unlikely race for the other thread to win, since it would usually require obtaining a large new allocation from the kernel while this thread is descheduled in a tight set of INSNs, or when there is an atomic modification while all the relevant INSNs are almost certainly already in the pipeline doing speculative work.    Also this is complicated by stashing the instantaneous value of segkmem_total_mem_allocated into a local variable (total_mem_used).

The best fix is probably to change the order of the tests, checking for > 75 first else test for > 70.

The simplest fix is to simply remove the "else" from the "} else if" line.

Alternatively, we could simply delete the entire else if branch, which is probably the second best approach.

Any of the above fixes other than deleting the else if branch would make a significant difference to the ARC dynamics when system memory is tight (and ARC is larger than arc_min), but whether that difference is better or worse is debatable.

```
                        // memory footprint has gotten really big, decrease spl_free substantially
                        int64_t total_mem_used = (int64_t) segkmem_total_mem_allocated;
                        if ((segkmem_total_mem_allocated * 100LL / real_total_memory) > 70) {
                                new_spl_free -= total_mem_used / 64;
                        } else if ((segkmem_total_mem_allocated * 100LL / real_total_memory) > 75) {
                                new_spl_free -= total_mem_used / 32;
                                lowmem = true;
                        }

rottegift:
(IIRC you read this stuff in email, @lundman, but I think the email will only show the initial accidental entry of the web form, and not it's update so check back to the web at https://github.com/openzfsonwindows/ZFSin/pull/32 :) )

PR #38, opened 2018-05-30
make FreeBSD system zpools importable 

jheuking:
FreeBSD system zpools were ignored on import, due to the unsupported partition type "FreeBSD ZFS". Furthermore the partition label had to start with "zfs". With this changes the partition label has to include "zfs" or "ZFS", which is the case for FreeBSD system zpools, which are labeled with "freebsd-zfs" by default.


michaeldexter:
Like!

Issue #43, opened 2018-06-06
copying or moving files fails with error "More data is available"


jheuking:
Copying or moving data from a zpool fails with error "More data is available".
The same error occurs when trying to read a directory with the command shell.

![copy_failed](https://user-images.githubusercontent.com/39560536/41023157-e091ce5a-696b-11e8-9d70-5d269a996094.PNG)

![read_failed](https://user-images.githubusercontent.com/39560536/41023558-0ad08c28-696d-11e8-9649-f785028bfb7c.PNG)

PR #45, opened 2018-06-11
remove falsely returned STATUS_BUFFER_OVERFLOW

jheuking:
possible fix for #43 

I don't know what this query was intended for, but as it is, it returns STATUS_BUFFER_OVERFLOW whenever the name of the specific dataset (e.g. "tank/ROOT/default") is longer than "NTFS" (at most) which seems not very helpful. 
This change even leads to the correct display of the volume label in the explorer.


jheuking:
Maybe it could be useful to add an assertion "space > name.length" at this point? 
Because if this is not the case, only part of the "NTFS"-string can be copied, which might lead to other errors... (unknown filesystem "NT" ☺)
But I don't know if this can even occur?

PR #48, opened 2018-06-22
return access denied when deleting files 

jheuking:
When deleting files with z_pflags ZFS_IMMUTABLE or ZFS_NOUNLINK set, zfs_zaccess_delete() returns EPERM and aborts. This function is called within IRP_MJ_CLEANUP though, which is too late to return STATUS_ACCESS_DENIED, so this has to be checked earlier, when v_unlink is set. The user now gets an error message when trying to delete such files.

### DIFF
--- a/ZFSin/spl/module/spl/spl-vnode.c
+++ b/ZFSin/spl/module/spl/spl-vnode.c
@@ -33,6 +33,7 @@
 //#include <IOKit/IOLib.h>
 
 #include <sys/taskq.h>
+#include <sys/zfs_znode.h>
 
 /* Counter for unique vnode ID */
 static uint64_t vnode_vid_counter = 0;
@@ -925,8 +926,14 @@ int   vnode_unlink(vnode_t *vp)
 
 void   vnode_setunlink(vnode_t *vp)
 {
+	if (vp && VTOZ(vp) && VTOZ(vp)->z_pflags & (ZFS_IMMUTABLE | ZFS_NOUNLINK)) {
+		dprintf("vnode_setunlink: access denied");
+		return STATUS_ACCESS_DENIED;
+	}	
 	dprintf("%s: \n", __func__);
+
 	vp->v_unlink = 1;
+	return STATUS_SUCCESS;
 }
 
 void *vnode_sectionpointer(vnode_t *vp)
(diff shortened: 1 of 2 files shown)


ZFSin/spl/module/spl/spl-vnode.c
```diff
@@ -925,8 +926,14 @@ int   vnode_unlink(vnode_t *vp)
 
 void   vnode_setunlink(vnode_t *vp)
 {
+	if (vp && VTOZ(vp) && VTOZ(vp)->z_pflags & (ZFS_IMMUTABLE | ZFS_NOUNLINK)) {
```
lundman: Generally, there should be no ZFS logic in SPL, so we might need to move this test up into ZFS, and only call setunlink when we decide to do so.

lundman: Since we call it so often, maybe make a zfs_setunlink() that does the test before calling vnode_setunlink.

jheuking: I'll give it a try. ☺
I just had another thought: Wouldn't it be better to just call zfs_zaccess_delete() directly in order to determine whether we are authorized to delete something? 
This is what the function is intended for, isn't it?

lundman: I honestly dont know, do we call it? not sure I even call it in osx version

jheuking: it gets called via fsdispatcher->IRP_MJ_CLEANUP->delete_entry->zfs_remove or ->zfs_rmdir in zfs_vnops.c like this:
```
if ((error = zfs_zaccess_delete(dzp, zp, cr))) {
	goto out;
}
```
this led me to the error initially. But I think, instead of picking single conditions out of zfs_zaccess_delete() and getting the same error again if any other condition is satisfied, it would be better to just use the function itself to determine whether we are going to run into access denied in IRP_MJ_CLEANUP...

lundman:
When I saw the ZFS conditional you added to `vnode_setunlink` in SPL, I just thought we could change all the ` vnode_setunlink` in ZFS to `zfs_setunlink()` and have the conditional there instead, and it calls `vnode_setunlink` when required, or return the error.

But if you want to check out `zfs_zaccess_delete` and use that for the test, that is fine by me :) I was just trying to avoid having ZFS in SPL, even though I am guilty of it with `zfs_vnop_recycle` callback I'll fix one day, honest...

jheuking:
reworked in  #49

PR #49, opened 2018-06-25
add zfs_setunlink() to check access rights before calling vnode_setunlink()

jheuking:
follow up #48

### DIFF
--- a/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
+++ b/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
@@ -341,6 +341,88 @@ int zfs_find_dvp_vp(zfsvfs_t *zfsvfs, char *filename, int finalpartmaynotexist, 
 }
 
 
+/*
+* call vnode_setunlink if zfs_zaccess_delete() allows it
+* TODO: provide credentials 
+*/
+
+NTSTATUS zfs_setunlink(vnode_t* vp, vnode_t* dvp) {
+
+	NTSTATUS Status = STATUS_UNSUCCESSFUL;
+
+	if (vp == NULL) {
+		Status = STATUS_INVALID_PARAMETER;
+		goto err;
+	}
+		
+
+	znode_t* zp = NULL;
+	znode_t* dzp = NULL;
+	zfsvfs_t *zfsvfs;
+	VN_HOLD(vp);
+	zp = VTOZ(vp);
+
+	if (vp && zp)
+		zfsvfs = zp->z_zfsvfs;
+	else {
+		Status = STATUS_INVALID_PARAMETER;
+		goto err;
+	}
+
+	/*
+		if dvp == null, find it
+	*/
+	if (dvp == NULL) {
+		uint64_t parent;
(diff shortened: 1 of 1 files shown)


ZFSin/zfs/module/zfs/zfs_vnops_windows.c
```diff
@@ -341,6 +341,88 @@ int zfs_find_dvp_vp(zfsvfs_t *zfsvfs, char *filename, int finalpartmaynotexist,
...
+	/*
+		if dvp == null, find it
+	*/
+	if (dvp == NULL) {
+		uint64_t parent;
+		if (sa_lookup(zp->z_sa_hdl, SA_ZPL_PARENT(zfsvfs),
+			&parent, sizeof(parent)) != 0) {
+			goto err;
+		}
+		if (zfs_zget(zfsvfs, parent, &dzp)) {
```
lundman: `zget` will automatically do a HOLD on the dvp...

ZFSin/zfs/module/zfs/zfs_vnops_windows.c
```diff
@@ -341,6 +341,88 @@ int zfs_find_dvp_vp(zfsvfs_t *zfsvfs, char *filename, int finalpartmaynotexist,
...
+	else {
+		dzp = VTOZ(dvp);
+	}
+
+	if (!dvp) {
+		Status = STATUS_NO_SUCH_FILE;
+		goto err;
+	}
+		
+	VN_HOLD(dvp);
```
lundman: So this HOLD should be moved into the else that doesnt call `zget` at line 388.

ZFSin/zfs/module/zfs/zfs_vnops_windows.c
```diff
@@ -341,6 +341,88 @@ int zfs_find_dvp_vp(zfsvfs_t *zfsvfs, char *filename, int finalpartmaynotexist,
...
+
+	int error = zfs_zaccess_delete(dzp, zp, 0);
+
+	if (error == 0) {
+		vnode_setunlink(vp);
+		Status = STATUS_SUCCESS;
+	}
+	else
+	{
```
lundman: I guess we should decide on the coding style too, either stay with ZFS like cstyle.pl, or allow VC++ style too

jheuking:
reworked in #50

ZFSin/zfs/module/zfs/zfs_vnops_windows.c
jheuking: I think staying with cstyle.pl would be ok. But in that case the script and configuration for checking the code style should at least be included in this repository...and maybe even better as a pre-commit hook?

lundman: Yes, cstyle.pl should be included, and is in osx. Just not sure how to run perl in windows yet :)

ZFSin/zfs/module/zfs/zfs_vnops_windows.c
jheuking: oh ok? I wondered about that, but i found the acquiring of holds in zfs_zget was indeed somewhat confusing...

PR #50, opened 2018-06-26
add zfs_setunlink() to check access rights before calling vnode_setunlink()

jheuking:
follow up #49 
made some rearrangements according to cstyle.pl and moved VN_HOLD() to the right position.

### DIFF
--- a/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
+++ b/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
@@ -342,6 +342,79 @@ int zfs_find_dvp_vp(zfsvfs_t *zfsvfs, char *filename, int finalpartmaynotexist, 
 
 
 
+/*
+ * Call vnode_setunlink if zfs_zaccess_delete() allows it
+ * TODO: provide credentials
+ */
+NTSTATUS zfs_setunlink(vnode_t *vp, vnode_t *dvp) {
+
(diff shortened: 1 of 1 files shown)


ZFSin/zfs/module/zfs/zfs_vnops_windows.c
```diff
@@ -342,6 +342,86 @@ int zfs_find_dvp_vp(zfsvfs_t *zfsvfs, char *filename, int finalpartmaynotexist,
...
+			dvp = NULL;
+			goto err;
+		}
+		dvp = ZTOV(dzp);
+	} else {
+		dzp = VTOZ(dvp);
+		VN_HOLD(dvp);
+	}
+
+	if (!dvp) {
```
lundman: Probably not needed, since we deal with == NULL case above...

but, you dont need to make a new PR each time, you can make changes, `git commit --amend` it then `git push -f` to force push over the commits in the PR.

You dont need to fix this, unless you want to practice :)

jheuking: Now that you mention it, this is maybe a bit over-cautious....but some git practice sounds always good, give me a sec ;)

jheuking: ah, seems like it worked :) :+1:

PR #52, opened 2018-07-10
show ioctl code as string

jheuking:
got tired of constantly looking up hex numbers...


lundman:
I like it, i think in future we might even be able to do some CPP macro magic and define the ioctl in one place, with string name!

PR #57, opened 2018-08-14
VN_HOLD/VN_RELE cleanup, INF fix for WDK 10.0.17134.0

jheuking:
50d1c35
When the driver takes the code path somewhere via FLUSH_BUFFERS -> zil_commit -> zil_commit_writer -> zfs_get_done(), zgd->zgd_rl was always NULL for me. Therefore it crashed in zfs_range_unlock(). So I guess this has to be checked before zfs_range_unlock() gets called. But I'm not sure if this should happen at all or if there is something else wrong?

9b37c32
I updated my Visual Studio/Windows 10 SDK/WDK from version 10.0.16299.15 to 10.0.17134.0 and after that I got an error that says that you can not have a DefaultInstall section alongside with a Manufacturer section... therefore I commented the DefaultInstall section out, as it shouldn't be needed anyways... 
Microsoft Docs:  [Note If you are building a universal or mobile driver package, this section is not valid and should not be used. Instead, use only the INF Manufacturer Section. Using both DefaultInstall and Manufacturer sections in your INF will cause Universal INF validation failures and can lead to inconsistent installation behaviors. ](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-defaultinstall-section)

ed24ca1
Some v_iocount fixes...

### DIFF
--- a/ZFSin/ZFSin.inf
+++ b/ZFSin/ZFSin.inf
@@ -22,15 +22,15 @@ ZFSin.DllFiles          = 11            ;%windir%\system32
 ;; Default install sections
 ;;
 
-[DefaultInstall]
-OptionDesc  = %ServiceDescription%
-CopyFiles   = ZFSin.DriverFiles
-;;,ZFSin.DllFiles
-RegisterDlls = shellzfsin
-CopyINF     = ZFSin.inf
-
-[DefaultInstall.Services]
-AddService  = %ServiceName%,0x802,ZFSin.Service
+;[DefaultInstall]
+;OptionDesc  = %ServiceDescription%
+;CopyFiles   = ZFSin.DriverFiles
+;;;,ZFSin.DllFiles
+;RegisterDlls = shellzfsin
+;CopyINF     = ZFSin.inf
+;
+;[DefaultInstall.Services]
+;AddService  = %ServiceName%,0x802,ZFSin.Service
 
 [Manufacturer]
 %Me%=Standard,NTamd64,NTx86
--- a/ZFSin/zfs/module/zfs/zfs_vnops.c
+++ b/ZFSin/zfs/module/zfs/zfs_vnops.c
@@ -1196,10 +1196,13 @@ zfs_get_done(zgd_t *zgd, int error)
 	objset_t *os = zp->z_zfsvfs->z_os;
 #endif
 
+	ASSERT(zgd->zgd_rl != NULL);
+
 	if (zgd->zgd_db)
 		dmu_buf_rele(zgd->zgd_db, zgd);
 
-	zfs_range_unlock(zgd->zgd_rl);
+	if (zgd->zgd_rl)
+		zfs_range_unlock(zgd->zgd_rl);
 
 	/*
 	 * Release the vnode asynchronously as we currently have the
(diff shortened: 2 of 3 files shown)


lundman:
https://github.com/openzfsonwindows/ZFSin/commit/50d1c358da8eec5b0bbfaf0c077e1481b9835349
It's supposed to be grabbed here, then passed along as `rl`.
https://github.com/openzfsonwindows/ZFSin/blob/master/ZFSin/zfs/module/zfs/zil.c#L1564

and then assigned over to zgl_rl in here somewhere
https://github.com/openzfsonwindows/ZFSin/blob/50d1c358da8eec5b0bbfaf0c077e1481b9835349/ZFSin/zfs/module/zfs/zfs_vnops.c#L1278

which looks to be missing.. tsk

lundman:
Oh forgot the compliment part, well done spotting these issues!

lundman:
_oh yeah, it's right there in zfs_get_data...overlooked that, shame on me_

and I missed the code to set `rl` :)

jheuking:
just have taken out the `#ifndef _WIN32` and it just works? 
wonder why this has been taken out in the first place
must be a relict of ancient times maybe..?
I left the zgd->zgd_rl != NULL verification there to avoid a BSOD, but also left an assertion for non privy folks, if something goes wrong in the future...

lundman:
The `zget` and `rl` lock was taken out of the `get_data` and moved into zil, so both are grabbed before we lock, to avoid a deadlock.  Now this is definitely needed on OsX, and I kept the code the same (which is why `get_data` has WIN32 tests) but I don't recall if I have had a deadlock or not.

jheuking:
ah, I think I found it in zil_lwb_commit(). There the range is locked and zfs_get_data is called with
```
#if defined (_WIN32) && defined (_KERNEL)
		error = zilog->zl_get_data(itx->itx_private, lrwb, dbuf, lwb, lwb->lwb_write_zio,
					zp, rl);
```

so we probably should take the rl, that is provided as argument in zfs_get_data instead of creating a new lock, shouldn't we?
somehow like this 
```
if (buf != NULL) { /* immediate write */

		zgd->zgd_rl = rl; //zfs_range_lock(zp, offset, size, RL_READER);
```

lundman:
Yep precisely that!

lundman:
Probably here:  https://github.com/openzfsonwindows/ZFSin/blob/d1f7148b4b9d26bbb85895727fcbd448333fc9df/ZFSin/zfs/module/zfs/zfs_vnops.c#L1269

jheuking:
probably the easiest way

and the zfs_range_unlock() here should be removed, shouldn't it?

https://github.com/openzfsonwindows/ZFSin/blob/d1f7148b4b9d26bbb85895727fcbd448333fc9df/ZFSin/zfs/module/zfs/zfs_vnops.c#L1308

lundman:
Yeah, should have been #ifndef'ed out - looks like I did that code on Friday afternoon eh

jheuking:
at least we are making progress :blush:
did we forgot something else? And I got you right, that you want all these changes to be 

> #ifndef'ed

right? :smile:

lundman:
It tends to help with future merges with upstream if its clear how it used to be :)

PR #76, opened 2018-10-22
should acquire spin lock when accessing vp

jheuking:
after the n-th test I got a crash in vnode_pager_setsize, probably [here](https://github.com/jheuking/ZFSin/blob/3b5d24439bef0e9df10fc89caab81746394bdaae/ZFSin/spl/include/sys/vnode.h#L434).
At this point the debugger claimed that the PFILE_OBJECT fileObject, which is defined inside the macro, points to something that looks like a valid FILE_OBJECT, but vp->fileobject was NULL. 
So I would suspect, that somehow `vnode_decouplefileobject()` was called in a different thread simultaneously. 
Of course this shouldn't happen, so we should grab a lock while we are accessing vp.

### DIFF
--- a/ZFSin/spl/include/sys/vnode.h
+++ b/ZFSin/spl/include/sys/vnode.h
@@ -416,6 +416,8 @@ static inline int win_has_cached_data(struct vnode *vp)
 	} while(0)
 #else
 #define vnode_pager_setsize(vp, sz)  do { \
+		KIRQL OldIrql; \
+		KeAcquireSpinLock(&vp->v_spinlock, &OldIrql); \
 		vp->FileHeader.AllocationSize.QuadPart = P2ROUNDUP((sz), PAGE_SIZE); \
 		vp->FileHeader.FileSize.QuadPart = (sz); \
 		vp->FileHeader.ValidDataLength.QuadPart = (sz); \
@@ -431,6 +433,7 @@ static inline int win_has_cached_data(struct vnode *vp)
 			} \
 			ObDereferenceObject(fileObject); \
 		} \
+		KeReleaseSpinLock(&vp->v_spinlock, OldIrql); \
 	} while(0)
 #endif
 
--- a/ZFSin/spl/module/spl/spl-vnode.c
+++ b/ZFSin/spl/module/spl/spl-vnode.c
@@ -1076,16 +1076,25 @@ FILE_OBJECT *vnode_fileobject(vnode_t *vp)
 }
 
 void vnode_couplefileobject(vnode_t *vp, FILE_OBJECT *fileobject) {
+	KIRQL OldIrql;
 	
 	if (fileobject)
 		fileobject->FsContext = vp;
-
-	vnode_setfileobject(vp, fileobject);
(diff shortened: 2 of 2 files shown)


ZFSin/spl/include/sys/vnode.h
```diff
@@ -416,6 +416,8 @@ static inline int win_has_cached_data(struct vnode *vp)
 	} while(0)
 #else
 #define vnode_pager_setsize(vp, sz)  do { \
+		KIRQL OldIrql; \
+		KeAcquireSpinLock(&vp->v_spinlock, &OldIrql); \
```
lundman: Rather than accessing internals of vnode, it should probably vnode_get(vp), to grab an iocount, test return code for success, and VN_RELE when done with it. Inside vnode_get we do the spinlock and make sure it isn't released.

ZFSin/spl/module/spl/spl-vnode.c
```diff
@@ -1076,16 +1076,25 @@ FILE_OBJECT *vnode_fileobject(vnode_t *vp)
 }
 
 void vnode_couplefileobject(vnode_t *vp, FILE_OBJECT *fileobject) {
+	KIRQL OldIrql;
 	
 	if (fileobject)
 		fileobject->FsContext = vp;
-
-	vnode_setfileobject(vp, fileobject);
+	if (vp) {
```
lundman: Similarly here, although, I think it would be acceptable to require vp to already be held when calling `vnode_couplefileobject()` ? It really should already be, one hopes :)

lundman:
This has been rolling around in my head a bit - there should be no call to `vnode_pager_setsize()` where it hasn't already grabbed an iocount - if there is, then that is wrong. But clearly vp->fileobject was "removed" even though `iocount > 0` which is the problem really. The vnode should not be freed/cleared while iocount is held.

But in this case, we presumably clear vp->fileobject because Windows told us to, so we "have" to.

I can see why you picked this solution - perhaps we need a vnode_lock()/unlock API to make it a bit nicer instead of accessing direct members (spinlock)

lundman:
A bit back in time, I thought that we could set vp->fileobject when enter dispatch() and have it removed when finishing. Naturally, that does not work when many threads are all called at the same time (one hopes they all have the same FileObject ptr at that?).

But there is an additional complication, the read-ahead, and lazy-write, magic, will also kick in by CCMgr. When we get MJ_CREATE to open a file, it will do the async call to "lock" the vp, and call MJ_READ from the read-ahead thread.  Meanwhile, the "original" call to MJ_CREATE can finish, and send MJ_CLEANUP.  The async MJ_READ need to be able to handle this, and I believe it does now. But would suggest there are cases where CLEANUP has been called, but a READ/WRITE request can come in after. (Which presumably we fail it).

Although, perhaps going back to setting fileobject only from dispatch entry, and clear on return could work if we used the TSD logic. That could actually work, spl_tsd() already exists.

Another option we could do is always pass fileobject down with vnode - but that is a lot of changes in ZFS that I would rather avoid :)

PR #77, opened 2018-10-24
GetFileDriveSize() needs function prototype

jheuking:
The compiler defaults to `int GetFileDriveSize();` as function declaration, because it cannot find the function while building. Afterwards, the linker is able to find something like GetFileDriveSize, but when the function returns LONGLONG and the compiler assumes int, the result is only rubbish of course.
Unfortunately several (or all?) libX projects define error level 1... so the compiler doesn't warn about such mistakes.
Noticed the error on a virtual hard disk with *exactly* 16Gb space. In that case the queried drive size was 0x400000000, but after the function returned, the value was 0.


lundman:
Amusingly, I just started looking at clearing out some warnings - since I can not currently run two Win10 VMs (upgraded to Fusion 11 which is a bit broken).

Incidentally, I think I awarded you direct repo access, so you don't have to do PR on things like that (that don't need review or discussion). But is fine if you want to continue as-is of course :)

jheuking:
oh really? If so, then I will try that with the next small changes :)

Issue #79, opened 2018-10-25
crash in spa_taskq_dispatch_ent()


jheuking:
After the recent commit 064ee8e8c1b877e816c1f4cc2357d192995cb526 I get a crash in  [`spa_taskq_dispatch_ent`](https://github.com/openzfsonwindows/ZFSin/blob/064ee8e8c1b877e816c1f4cc2357d192995cb526/ZFSin/zfs/module/zfs/spa.c#L1039) where `&spa->spa_zio_taskq` is completely uninitialized (everything NULL).
When it gets accessed [here](https://github.com/openzfsonwindows/ZFSin/blob/064ee8e8c1b877e816c1f4cc2357d192995cb526/ZFSin/zfs/module/zfs/spa.c#L1048).. boom!

jheuking:
guess this is solved then 👍

PR #80, opened 2018-10-25
spa_create_zio_taskqs: just do it

jheuking:
regarding #79 
tried to directly push it to master, but unfortunately got error 403, so yeah... here I am again 😬


lundman:
hah, whoops - my bad. That `p0` has been a thorn even since osx start. But I changed my mind halfway through this work, it should be `== NULL` instead, and we'll use NULL to signify the first `p0` process.

I fiddled with the membership permissions, perhaps you can try a direct commit now?

jheuking:
hey! it even works directly from github, thanks :smile: 👍

lundman:
Perfect!

Issue #78, opened 2018-10-25
There is some label offset miscalculation


jheuking:
As we now get the right size from `GetFileDriveSize();` we can also read the vdev label properly ... 🎉  
but that revealed some problems with the label offset calculation and in addition to that we create configs now that we shouldn't create.

1. Currently we read the partition labels at the end of the partition at partition offset + drivesize, which is far behind the disk in most cases. So we never find anything valid there. 

2. We are now able to find the partition label at the end of the disk, which leads to a problem. 
Starting from a device that has one ZFS partition with a zpool we basically go through the devices like that:
```
foreach dev that SetupDiEnumDeviceInterfaces throws at us{
add as potential device to avltree slice_cache

does this device has partitions that could contain a zpool? 
YES? Then add the device again to slice_cache with partition information hacked into dev name
}

ok, we have some configs
now check which could be ZFS pools, i.e. read the vdev label and do some other checking..
```
and now comes the cool stuff
as we found valid vdev label on the whole disk AND on the partition we take both configuration as valid.
I don't know if this is the way it should work, but I guess it's not :smile:
the problem occurs here, because we have a valid configuration (from the whole disk) that doesn't contain the partition offset hack, so the driver aborts afterwards.
Before we added the fix for GetFileDriveSize() there weren't able to read any label for the whole disk so that was ignored and everything worked as expected..
So yeah... I'm not sure how to solve this the best way.

But I would suggest to just allow configs, which include the hack in the device path. That would currently limit us to only use zpools on partitions. But importing a zpool that is directly embedded in a disk, crashed the driver anyways in my case... 

update: ...had a little mistake in reading the drivesize, now the zpool can be read directly from the disk also
so maybe we should check if the disk has partitions and if so, search the partitions for a zpool and otherwise check the disk itself for a zpool

PR #81, opened 2018-10-25
either search zpool on the disk itself OR within partitions on the disk

jheuking:
regarding #78 
now we should be able to import pools again, whether they are located directly on disk or within a partition...
fix for reading the vdev label at the end of a partition still missing...


lundman:
Ah awesome - I did thing there was a bug in this area but it eluded me. Also, do not spend too much time on the back labels - I think the math misses on OSX too, something peculiar when code came from illumos to ZOL then us. Might need checking illumos work again (and perhaps, making sure a test pool is created by illumos, since we probably create them badly.

PR #82, opened 2018-10-29
ZFS installer

jheuking:
At the moment the installer is able make a clean install and uninstall of our driver.
It calls InstallHinfSection to install and uninstall (like we would do with rundll32 from the commandline). Additionally, on uninstall, it send ZFS_IOC_UNREGISTER_FS to our driver, before it calls InstallHinfSection, so we have the chance to clean everything before we get unloaded.
I took DefaultInstall and DefaultUninstall back in because it makes the Installation a lot easier. 
I haven't figured out how to install the driver through the Manufacturer-section without using devcon. So if we would use devcon for installation, it would create a device node for our driver (ZFSin Controller or something?) which we don't make use of. But the device node needs to be deleted to unload the driver. I think this is just extra work, that doesn't lead to any advantages in my opinion...
So I switched to the DefaultInstall-sections again..
Additionally there were some problems in spl_stop(), which I tried to get along with in 
05d3c26.
Install/Uninstall like that:
`zfsinstaller install ZFSin\ZFSin.inf`
`zfsinstaller uninstall ZFSin\ZFSin.inf`

For debugging, I added `sc query ZFSin` before and after each operation. 
On Installation the service should first not be found and afterwards should show the state RUNNING.
On Uninstall the service should first be in state RUNNING and afterwards be deleted. 
If there are problems on unloading (open handles, etc.) the service will show STOP_PENDING after the uninstall. 

What still bothers me a bit is the send_zfs_ioc_unregister_fs function in the installer, because I had to define a few things, that should be read from zfs_ioctl.h actually.
But we cannot just include that header. Maybe we should split that into zfs_ioctl_def.h and zfs_ioctl.h to make definitions (like `#define ZFSIOCTL_TYPE 40000`) in zfs_ioctl_def.h, which is includable in the installer.


ZFSin/spl/module/spl/spl-vmem.c
```diff
@@ -2139,9 +2139,9 @@ vmem_destroy(vmem_t *vmp)
 
 	kstat_delete(vmp->vm_ksp);
 
+	vmem_free(vmem_vmem_arena, vmp, sizeof (vmem_t));
 	mutex_destroy(&vmp->vm_lock);
 	cv_destroy(&vmp->vm_cv);
-	vmem_free(vmem_vmem_arena, vmp, sizeof (vmem_t));
```
lundman: Isn't that freeing the memory of `vmp` then dereferencing it in `vmp->vm_lock` ?

ZFSin/spl/include/sys/systm.h
```diff
@@ -39,7 +39,9 @@ static inline void bsd_untimeout(void(*func)(void *), void *ID)
 	struct bsd_timeout_wrapper *btw = (struct bsd_timeout_wrapper *)ID;
 	LARGE_INTEGER p = { -1 };
 	btw->init = 0;
-	KeSetTimer(&btw->timer, p, NULL);
+	ASSERT(btw->init == BSD_TIMEOUT_MAGIC); // timer was not initialized 
+	if(btw->init == BSD_TIMEOUT_MAGIC)
+		KeSetTimer(&btw->timer, p, NULL);
```
lundman: I wrote the bsdtimeout code and then found I couldn't unload the driver to test unloading it, so I'm not surprised the unload needs fixes. But I think I'll try to find a way that doesn't use the "debug build" init magic value. I'll put it on my todo for a quiet day :)   (Especially since we set `init` to 0 just before the ASSERT

ZFSin/zfs/include/sys/zfs_ioctl.h
```diff
@@ -598,8 +598,9 @@ typedef enum zfs_ioc {
 	//ZFS_IOC_WINDOWS= ('Z' << 8) + 0xE0,
 	ZFS_IOC_MOUNT				= CTL_CODE(ZFSIOCTL_TYPE, 0x8E0, METHOD_NEITHER, FILE_ANY_ACCESS),
 	ZFS_IOC_UNMOUNT				= CTL_CODE(ZFSIOCTL_TYPE, 0x8E1, METHOD_NEITHER, FILE_ANY_ACCESS),
+	ZFS_IOC_UNREGISTER_FS		= CTL_CODE(ZFSIOCTL_TYPE, 0x8E2, METHOD_NEITHER, FILE_ANY_ACCESS),
 
-	ZFS_IOC_LAST				= CTL_CODE(ZFSIOCTL_TYPE, 0x8E2, METHOD_NEITHER, FILE_ANY_ACCESS),
+	ZFS_IOC_LAST				= CTL_CODE(ZFSIOCTL_TYPE, 0x8E3, METHOD_NEITHER, FILE_ANY_ACCESS),
 } zfs_ioc_t;
```
lundman: Technically, the IOCTLs actually live in `include/sys/fs/zfs.h` on illumos, I had to move them into `zfs_ioctl.h` for OSX. So if we want to break them out into its own file and just include, that is fine.

ZFSin/zfs/module/zfs/zfs_ioctl.c
```diff
@@ -5909,6 +5909,18 @@ zfs_ioctl_register_dataset_modify(zfs_ioc_t ioc, zfs_ioc_legacy_func_t *func,
 							  DATASET_NAME, B_TRUE, POOL_CHECK_SUSPENDED | POOL_CHECK_READONLY);
 }
 
+int
+zfs_ioc_unregister_fs(void) {
+	//IoUnregisterFsRegistrationChange(WIN_DriverObject, DriverNotificationRoutine);
+	dprintf("%s\n", __func__);
+	IoUnregisterFileSystem(fsDiskDeviceObject);
+	IoDeleteDevice(fsDiskDeviceObject);
+	IoDeleteDevice(ioctlDeviceObject);
```
lundman: Don't we create 3 devices? disk, fs and ioctl? Should we have the 3rd here for completeness?

lundman:
This is looking great, I look forward to trying it once vmware fixes their stuff :)

ZFSin/spl/module/spl/spl-vmem.c
jheuking: oh yeah, you could be right. But never got a crash there though 😏. Should take a closer look and change that back maybe

ZFSin/spl/include/sys/systm.h
jheuking: oh, just had one of that "what the heck" moments 😂 Guess I just overlooked that `btw->init = 0;` the first time, but I would say leaving an assertion as a reminder here doesn't hurt though ...

ZFSin/zfs/module/zfs/zfs_ioctl.c
jheuking: I think the diskDeviceObject what you are referencing is created per dataset in `zfs_windows_mount`, isn't it? In `zfs_attach` the `fsDiskDeviceObject` and `ioctlDeviceObject` are the only devices that are created (If I haven't overlooked anything again :))

ZFSin/zfs/include/sys/zfs_ioctl.h
jheuking: ok, thats good. More work for times when everything else is overly frustrating again :smile:

jheuking:
so basically I would take a closer look at the part with vmem_free() before merging that into master 
everything else could be improved over time step by step, my suggestion

ZFSin/spl/module/spl/spl-vmem.c
jheuking: hm, so as it seems vmem_xfree gets called from here most times... therefore we don't have to deal with freed memory. But it also works the other way round for me, so maybe just leave that as it is for the moment

PR #85, opened 2018-11-09
do manual mount if mountmgr refuses

jheuking:
this should improve #5 
haven't figured out yet how to get the mountmgr to do everything automatically.
But manually removing and setting the mountpoints seems to work (at least for the explorer).
So now we generally remove all the mountpoints for our volume in unmount. 
When mounting, we check if the mountpoints were set appropriately after the volume arrival ioctl was sent.
If the mountpoints are missing, we create them manually.
Would be nice, if this could be reviewed. Not sure if the mountmgr has equal behavior on different systems😬

### DIFF
--- a/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
+++ b/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
@@ -1494,6 +1494,8 @@ NTSTATUS ioctl_disk_get_drive_geometry(PDEVICE_OBJECT DeviceObject, PIRP Irp, PI
 	}
 
 	zfsvfs_t *zfsvfs = vfs_fsprivate(zmo);
+	if (zfsvfs == NULL)
+		return STATUS_INVALID_PARAMETER;
 
 	ZFS_ENTER(zfsvfs);  // This returns EIO if fail
 
@@ -1538,6 +1540,8 @@ NTSTATUS ioctl_disk_get_drive_geometry_ex(PDEVICE_OBJECT DeviceObject, PIRP Irp,
 	}
 
 	zfsvfs_t *zfsvfs = vfs_fsprivate(zmo);
+	if (zfsvfs == NULL)
+		return STATUS_INVALID_PARAMETER;
 
 	ZFS_ENTER(zfsvfs);  // This returns EIO if fail
 
@@ -1583,6 +1587,8 @@ NTSTATUS ioctl_disk_get_partition_info(PDEVICE_OBJECT DeviceObject, PIRP Irp, PI
 	}
 
 	zfsvfs_t *zfsvfs = vfs_fsprivate(zmo);
+	if (zfsvfs == NULL)
+		return STATUS_INVALID_PARAMETER;
 
 	ZFS_ENTER(zfsvfs);  // This returns EIO if fail
 
@@ -1625,6 +1631,8 @@ NTSTATUS ioctl_disk_get_partition_info_ex(PDEVICE_OBJECT DeviceObject, PIRP Irp,
 	}
 
 	zfsvfs_t *zfsvfs = vfs_fsprivate(zmo);
+	if (zfsvfs == NULL) 
+		return STATUS_INVALID_PARAMETER;
 
 	ZFS_ENTER(zfsvfs);  // This returns EIO if fail
 
@@ -1668,6 +1676,8 @@ NTSTATUS ioctl_disk_get_length_info(PDEVICE_OBJECT DeviceObject, PIRP Irp, PIO_S
 	}
 
 	zfsvfs_t *zfsvfs = vfs_fsprivate(zmo);
+	if (zfsvfs == NULL)
+		return STATUS_INVALID_PARAMETER;
 
 	ZFS_ENTER(zfsvfs);  // This returns EIO if fail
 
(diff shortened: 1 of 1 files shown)


ZFSin/zfs/module/zfs/zfs_vnops_windows_lib.c
```diff
@@ -246,14 +246,46 @@ NTSTATUS mountmgr_add_drive_letter(PDEVICE_OBJECT mountmgr, PUNICODE_STRING devp
...
 }
 
-NTSTATUS mountmgr_get_drive_letter(PDEVICE_OBJECT mountmgr, PUNICODE_STRING devpath, char *savename)
+/*
+* check if valid mountpoint, like \DosDevices\X:
+*/
+BOOLEAN MOUNTMGR_IS_DRIVE_LETTER_A(char *mountpoint) {
+	UNICODE_STRING wc_mpt;
+	wchar_t buf[PATH_MAX];
+	mbstowcs(buf, mountpoint, 200);
```
lundman: nit: `200` - perhaps sizeof(buf) or sizeof(wchar_t)*PATH_MAX ?

ZFSin/zfs/module/zfs/zfs_vnops_windows_lib.c
```diff
@@ -246,14 +246,46 @@ NTSTATUS mountmgr_add_drive_letter(PDEVICE_OBJECT mountmgr, PUNICODE_STRING devp
...
+	return MOUNTMGR_IS_DRIVE_LETTER(&wc_mpt);
+}
+
+/*
+* check if valid mountpoint, like \??\Volume{abc}
+*/
+BOOLEAN MOUNTMGR_IS_VOLUME_NAME_A(char *mountpoint) {
+	UNICODE_STRING wc_mpt;
+	wchar_t buf[PATH_MAX];
+	mbstowcs(buf, mountpoint, 200);
```
lundman: ditto

ZFSin/zfs/module/zfs/zfs_vnops_windows_lib.c
```diff
@@ -836,6 +916,128 @@ NTSTATUS CreateReparsePoint(POBJECT_ATTRIBUTES poa, LPCWSTR SubstituteName, LPCW
 	return status;
 }
 
+
+/*
+* go through all mointpoints (IOCTL_MOUNTMGR_QUERY_POINTS) and check if our driveletter is in the list
+* return 1 if yes, otherwise 0
+*/
+NTSTATUS mountmgr_is_driveletter_assigned(PDEVICE_OBJECT mountmgr, wchar_t driveletter, BOOLEAN *ret) {
```
lundman: nit: we should probably stay with OpenZFS `cstyle.pl` coding standard, even if I have also been a bit lax with it :) "{" on new line for functions, and all that.

ZFSin/zfs/module/zfs/zfs_vnops_windows_lib.c
```diff
@@ -836,6 +916,128 @@ NTSTATUS CreateReparsePoint(POBJECT_ATTRIBUTES poa, LPCWSTR SubstituteName, LPCW
...
+
+
+
+void generateGUID(char* pguid) {
+	char *uuid_format = "xxxxxxxx-xxxx-4xxx-Nxxx-xxxxxxxxxxxx";
+	char *szHex = "0123456789ABCDEF-";
+	int len = strlen(uuid_format);
+
+	for (int i = 0; i < len + 1; i++)
+	{
```
lundman: nit: more cstyle things

ZFSin/zfs/module/zfs/zfs_vnops_windows_lib.c
```diff
@@ -836,6 +916,128 @@ NTSTATUS CreateReparsePoint(POBJECT_ATTRIBUTES poa, LPCWSTR SubstituteName, LPCW
...
+				ipoint->DeviceNameLength / sizeof(WCHAR), DeviceName,
+				ipoint->SymbolicLinkNameLength / sizeof(WCHAR), SymbolicLinkName);
+
+			ULONG len = 0;
+			RtlUnicodeToUTF8N(mpt_name, MAXPATHLEN, &len, SymbolicLinkName, ipoint->SymbolicLinkNameLength);
+			mpt_name[len] = 0;
+			char c_driveletter;
+			wctomb(&c_driveletter, driveletter);
+			if (MOUNTMGR_IS_DRIVE_LETTER_A(mpt_name) && mpt_name[12] == c_driveletter) {
+				*ret = 1;
```
lundman: Do we know who/what releases the `kmem_alloc` above in this positive exit branch? Is it one of those "Windows will free it?" if so, we should probably not use kmem. If we are to free it, where do we?

jheuking: oh yeah, that should be freed before return 
will do that

ZFSin/zfs/module/zfs/zfs_vnops_windows_lib.c
jheuking: that's probably some copy-paste leftover... sizeof(buf) will do it, i guess :)

jheuking:
> With the mount things, there were things I had to test,

this changes should only affect mounting with driveletter though..

Issue #94, opened 2018-12-31
Partition based zpool is not imported


raindev:
I have a zpool residing on a partition of an external USB drive. The pool was created on Ubuntu 18.10 (`userobj_accounting` feature disabled), I can successfully import it with ZFS on macOS. When I try to import the zpool on Windows it fails with "no pool available to import". Below is the full output:

```
PS C:\WINDOWS\system32> zpool import -a
path '\\?\usbstor#disk&ven_toshiba&prod_external_usb_3.0&rev_5438#20170222008431f&0#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}'
 and '\\?\PhysicalDrive1'
read partitions ok 7
    gpt 0: type 6c54e010 off 0x100000 len 0xc800000
    gpt 1: type 6c54e010 off 0xc900000 len 0x3ff8000000
    gpt 2: type 6c54e010 off 0x4004900000 len 0x8000000
    gpt 3: type 6c54e010 off 0x4014900000 len 0x3ff8000000
    gpt 4: type 6c54e010 off 0x800c900000 len 0x8000000
    gpt 5: type 6c54e010 off 0x801c900000 len 0x4000000000
    gpt 6: type 6c54e010 off 0xc01c900000 len 0x10000000000
asking libefi to read label
EFI read OK, max partitions 128
    part 0:  offset 800:    len 64000:    tag: c    name: ''
    part 5:  offset 400e4800:    len 20000000:    tag: 11    name: ''
    part 6:  offset 600e4800:    len 80000000:    tag: 18    name: ''
path '\\?\scsi#disk&ven_nvme&prod_samsung_ssd_970#5&2b537fae&0&000000#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}'
 and '\\?\PhysicalDrive0'
read partitions ok 5
    gpt 0: type 6c54e010 off 0x100000 len 0x20000000
    gpt 1: type 6c54e010 off 0x20100000 len 0x3a35300000
    gpt 2: type 6c54e010 off 0x3a55400000 len 0x1000000
    gpt 3: type 6c54e010 off 0x3a56400000 len 0x39fc035e00
    gpt 4: type 6c54e010 off 0x7452500000 len 0x1e600000
asking libefi to read label
EFI read OK, max partitions 128
    part 0:  offset 800:    len 100000:    tag: c    name: 'EFI System Partition'
[...]

PR #95, opened 2018-12-31
Document partition label requirement in readme

raindev:
A partition label has to contain "zfs" for the pool to be importable.
Documents the issue I've run into in #94.


lundman:
Oh, I was about to merge, but alas, I went and implemented readonly imports.

raindev:
That's even better than merging. I've removed now irrelevant commit from the pull request.

PR #112, opened 2019-03-01
Document usage of `kstat`

zfs2019:
By default, ZFS eat up a lot of RAM. We need a way to set `zfs_arc_max`, and other settings, but there is no documentation about that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,3 +401,17 @@ No pools available
 ```
 
 A reboot might be neccessary to uninstall it completely.
+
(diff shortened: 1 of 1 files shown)


lundman:
Yes that should be added to the wiki for sure. There are some hints in https://github.com/openzfsonwindows/ZFSin/commit/d9800daec5e306c26057cc073686a5662191ad82

lundman:
Feel free to add something about it in Wiki.

PR #122, opened 2019-04-02
Add TRIM support

lundman:
UNMAP/TRIM support is a frequently-requested feature to help
prevent performance from degrading on SSDs and on various other
SAN-like storage back-ends.  By issuing UNMAP/TRIM commands for
sectors which are no longer allocated the underlying device can
often more efficiently manage itself.

This TRIM implementation is modeled on the `zpool initialize`
feature which writes a pattern to all unallocated space in the
pool.  The new `zpool trim` command uses the same vdev_xlate()
code to calculate what sectors are unallocated, the same per-
vdev TRIM thread model and locking, and the same basic CLI for
a consistent user experience.  The core difference is that
instead of writing a pattern it will issue UNMAP/TRIM commands
for those extents.

The zio pipeline was updated to accommodate this by adding a new
ZIO_TYPE_TRIM type and associated spa taskq.  This new type makes
is straight forward to add the platform specific TRIM/UNMAP calls
to vdev_disk.c and vdev_file.c.  These new ZIO_TYPE_TRIM zios are
handled largely the same way as ZIO_TYPE_READs or ZIO_TYPE_WRITEs.
This makes it possible to largely avoid changing the pipieline,
one exception is that TRIM zio's may exceed the 16M block size
limit since they contain no data.

In addition to the manual `zpool trim` command, a background
automatic TRIM was added and is controlled by the 'autotrim'
property.  It relies on the exact same infrastructure as the
manual TRIM.  However, instead of relying on the extents in a
metaslab's ms_allocatable range tree, a ms_trim tree is kept
per metaslab.  When 'autotrim=on', ranges added back to the
ms_allocatable tree are also added to the ms_free tree.  The
ms_free tree is then periodically consumed by an autotrim
thread which systematically walks a top level vdev's metaslabs.

Since the automatic TRIM will skip ranges it considers too small
there is value in occasionally running a full `zpool trim`.  This
may occur when the freed blocks are small and not enough time
was allowed to aggregate them.  An automatic TRIM and a manual
`zpool trim` may be run concurrently, in which case the automatic
TRIM will yield to the manual TRIM.

Reviewed-by: Jorgen Lundman <lundman@lundman.net>
Reviewed-by: Tim Chase <tim@chase2k.com>
Reviewed-by: Matt Ahrens <mahrens@delphix.com>
Reviewed-by: George Wilson <george.wilson@delphix.com>
Reviewed-by: Serapheim Dimitropoulos <serapheim@delphix.com>
Contributions-by: Saso Kiselkov <saso.kiselkov@nexenta.com>
Contributions-by: Tim Chase <tim@chase2k.com>
Contributions-by: Chunwei Chen <tuxoko@gmail.com>
Signed-off-by: Brian Behlendorf <behlendorf1@llnl.gov>

Get rid of space_map_update() for ms_synced_length

Initially, metaslabs and space maps used to be the same thing
in ZFS. Later, we started differentiating them by referring
to the space map as the on-disk state of the metaslab, making
the metaslab a higher-level concept that is metadata that deals
with space accounting. Today we've managed to split that code
furthermore, with the space map being its own on-disk data
structure used in areas of ZFS besides metaslabs (e.g. the
vdev-wide space maps used for zpool checkpoint or vdev removal
features).

This patch refactors the space map code to further split the
space map code from the metaslab code. It does so by getting
rid of the idea that the space map can have a different in-core
and on-disk length (sm_length vs smp_length) which is something
that is only used for the metaslab code, and other consumers
of space maps just have to deal with. Instead, this patch
introduces changes that move the old in-core length of the
metaslab's space map to the metaslab structure itself (see
ms_synced_length field) while making the space map code only
care about the actual space map's length on-disk.

The result of this is that space map consumers no longer have
to deal with syncing two different lengths for the same
structure (e.g. space_map_update() goes away) while metaslab
specific behavior stays within the metaslab code. Specifically,
the ms_synced_length field keeps track of the amount of data
metaslab_load() can read from the metaslab's space map while
working concurrently with metaslab_sync() that may be
appending to that same space map.

As a side note, the patch also adds a few comments around
the metaslab code documenting some assumptions and expected
behavior.

Reviewed-by: Matt Ahrens <mahrens@delphix.com>
Reviewed-by: Brian Behlendorf <behlendorf1@llnl.gov>
Reviewed by: Pavel Zakharov <pavel.zakharov@delphix.com>
Signed-off-by: Serapheim Dimitropoulos <serapheim@delphix.com>

Fix guid memory leak

When a pool is full the sync task may not run due to lack of space.
This can result in a memory leak when the sync task provied to
dsl_sync_task_nowait() is responsible for freeing the passed arg.
For now, handle this by switch to ZFS_SPACE_CHECK_NONE to ensure
it always gets called which is already what's done for the other
dsl_sync_task_nowait() callers.

A more complete fix has been provided for initial review as
https://github.com/zfsonlinux/zfs/pull/8497.

Signed-off-by: Brian Behlendorf <behlendorf1@llnl.gov>

This commits also removes repeated "channel_program" ioctl register and
fixes a bug in kstat dprintf.


lundman:
Tested both file and disk paths:

```
\??\C:\src\poolfile.bin  ONLINE       0     0     0  (100% trimmed, completed at 19/04/02 12:06:55 PM)
```
```
  pool: tank2
 state: ONLINE
  scan: none requested
config:

        NAME                   STATE     READ WRITE CKSUM
        tank2                  ONLINE       0     0     0
          /dev/physicaldrive2  ONLINE       0     0     0  (32% trimmed, started at 19/04/02 02:45:58 PM)
```

```
vdev_disk_open: nonrot 1, trim 1, securetrim 0
trim said 0x0

vdev_file_open: set Sparse 0x0.
filetrim said 0x0
```

```
$ ./kstat.exe | grep trim
name:   z_trim_iss                      class:    taskq
name:   z_trim_int                      class:    taskq
name:   z_trim_iss                      class:    taskq
name:   z_trim_int                      class:    taskq
        zfs_trim_extent_bytes_max       134217728
        zfs_trim_extent_bytes_min       32768
        zfs_trim_metaslab_skip          0
        zfs_trim_queue_limit            10
        zfs_trim_txg_batch              32
        autotrim_bytes_failed           0
        autotrim_bytes_skipped          0
        autotrim_bytes_written          0
        autotrim_extents_failed         0
        autotrim_extents_skipped        0
        autotrim_extents_written        0
        trim_bytes_failed               0
        trim_bytes_skipped              0
        trim_bytes_written              0
        trim_extents_failed             0
        trim_extents_skipped            0
        trim_extents_written            0
        autotrim_bytes_failed           0
        autotrim_bytes_skipped          0
        autotrim_bytes_written          0
        autotrim_extents_failed         0
        autotrim_extents_skipped        0
        autotrim_extents_written        0
        trim_bytes_failed               0
        trim_bytes_skipped              65536
        trim_bytes_written              1953587200
        trim_extents_failed             0
        trim_extents_skipped            6
        trim_extents_written            16
```

Prints will be removed before merge.

It also highlights that `vdev_file.c` is only compiled in kernel mode. Unlike `vdev_disk.c` which is only compiled in kernel mode, `vdev_file.c` is also compiled in userland. It is used for `zdb` which opens devices as `files` - including disks. Currently our `zdb` will be lacking until we add this functionality.

lundman:
The Windows specific code is;

Setting of the storage capabilities:
https://github.com/openzfsonwindows/ZFSin/pull/122/commits/4b5133fc7d60d428300fc46a306e6b0cc3e995cc#diff-58dbde9c750edd0211c8997d8c74f926R329

Calling TRIM on disk:
https://github.com/openzfsonwindows/ZFSin/pull/122/commits/4b5133fc7d60d428300fc46a306e6b0cc3e995cc#diff-58dbde9c750edd0211c8997d8c74f926R594

Implementations of API:
https://github.com/openzfsonwindows/ZFSin/pull/122/commits/4b5133fc7d60d428300fc46a306e6b0cc3e995cc#diff-21cd8b16bd9b9171f9c98386761f8034R222
(Which includes bug fix for kernel_ioctl)

Changing file based pools to sparse;
https://github.com/openzfsonwindows/ZFSin/pull/122/commits/4b5133fc7d60d428300fc46a306e6b0cc3e995cc#diff-c7547fb0a86ae6fe3b8d9f0c25e9fc4eR247

Trimming file;
https://github.com/openzfsonwindows/ZFSin/pull/122/commits/4b5133fc7d60d428300fc46a306e6b0cc3e995cc#diff-c7547fb0a86ae6fe3b8d9f0c25e9fc4eR446

Whether or not it is worth wrapping that in flock/VOP_SPACE just to change it into FILE_ZERO_DATA_INFROMATION;

https://github.com/openzfsonwindows/ZFSin/pull/122/commits/4b5133fc7d60d428300fc46a306e6b0cc3e995cc#diff-21cd8b16bd9b9171f9c98386761f8034R344

but if we ever have a second call to `VOP_SPACE` it might be.

lundman:
Merged with 9dd8832

PR #138, opened 2019-05-17
Mutex replacement

lundman:
Attempt to replace faulty fast_mutex usage with events. Additionally, try to avoid blocking if possible by using atomic CAS. 

Initial tests behave as expected, but let's make sure we get all the corner cases.


lundman:
Merged with master 0614315

PR #161, opened 2019-08-26
Reimplement REPARSE POINTS

lundman:
In an attempt to be compatible with existing reparse poinrts (EA stored) - for example samba implementation.

I believe this is not complete, but a starting point - there are also some cross over with master.

### DIFF
--- a/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
+++ b/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
@@ -119,6 +119,12 @@ uint64_t vnop_num_reclaims = 0;
 uint64_t vnop_num_vnodes = 0;
 #endif
 
+#pragma region Prototypes
+static void zp_unpack_times(znode_t *zp, PLARGE_INTEGER pMtime, PLARGE_INTEGER pCtime, PLARGE_INTEGER pCrtime, PLARGE_INTEGER pAtime);
+static NTSTATUS vnode_apply_eas(struct vnode *vp, PFILE_FULL_EA_INFORMATION ea, ULONG eaLength, PULONG eaErrorOffset);
+static BOOLEAN vattr_apply_single_ea(vattr_t *vap, PFILE_FULL_EA_INFORMATION ea);
+#pragma endregion
+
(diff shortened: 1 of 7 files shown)


ZFSin/zfs/module/zfs/zfs_vnops_windows.c
```diff
@@ -542,8 +543,7 @@ int zfs_find_dvp_vp(zfsvfs_t *zfsvfs, char *filename, int finalpartmaynotexist,
 	return 0;
 }
 
-static ULONG zp_get_reparse_tag(znode_t *zp);
-// There are multiple struct types with "stat" style members.
+// There are multiple sturct types with "stat" style members.
```
JMoVS: I think this is a typo, right? :D "sturct"

lundman:
Quite interesting - even knowing the names like `FsRtlGetNextExtraCreateParameter()` it is hard to find much information at all, and nothing at all for something like `GUID_ECP_CREATE_REDIRECT`. Sure it has a struct with it, but not much information as to when it should be triggered etc.

`GUID_ECP_QUERY_ON_CREATE` Is interesting as well, since master had different `file_stat_information()` to yours, it looks a bit different.

Is it acceptable then to store the reparse points in symlink data (`readlink()` etc)? I was under the impression that samba stores it in an EA - although I have been unable to get samba to create me a symlink on a test pool. 

I'll create a parallel PR to this, as your code gets merged into master - probably the easiest way for us to keep an eye on code changes.

lundman:
I'm getting increasingly more unhappy with `zfs_vnop_lookup()` - it has become large and unwieldy, and sensitive to change. I've been hesitate to touch it, since it would break for a bit, but I think I might have to just do it.

At the very least, wrap it in one more function, so we don't have to repeat the cleanup code so many times. It seems clear that QOC should be done for file and dirs, for existing and creating, so adding 4 copies of the same code is dumb. 

Anyway, head's up I'm about to break it :)

lundman:
https://github.com/openzfsonwindows/ZFSin/tree/rp_synth

PR #162, opened 2019-09-03
[WIP] Alternative build system using CMake

nazar554:
Might be useful for people who use QtCreator, or don't like `.vcxproj`

- Need to build userspace tools
- Figure out how to sign the driver
- Build an installer


nazar554:
The only thing left is bulilding the installer using InnoSetup.
Had a few problems with missing function prototypes (since I am building with `/we4013`), resolved them manually. `zpool` and `zstreamdump` also needed `libuuid`, needed to add missing include and project reference.

nazar554:
I'm a bit confused about this in `sys/types32.h`:
```
typedef uint32_t zsize_t;
#define size_t zsize_t

typedef uint64_t zoff_t;
#define off_t zoff_t
```

How the MSVC C standard library headers should continue to function with header included?
I am getting thousands of warnings about it in `printf` and pointer to size_t casts everywhere

lundman:
Ok, so the `size_t` workaround was a misunderstand by myself when I started. I had it in mind that it should be 32bit, and did that hack, plus hundreds of changes. It is on my todo to undo all that, and restore it to its former glory.

Even in devstudio there are thousands of warnings, I did notice you have a nice list of warning-disables which I might borrow, since a lot of the warnings I get are irrelevant.

Long-term though, what are your thoughts? Maintain two build systems? I went with devstudio on the assumption that most windows developers would be using that. (And I must admit that now that I have kernel debugger working in devstudio, it is rather nice.)

ZFSin/zfs/module/zfs/zfs_windows_zvol.c
```diff
@@ -892,7 +892,7 @@ wzvol_StopAdapter(__in pHW_HBA_EXT pHBAExt)               // Adapter device-obje
 #if defined(_AMD64_)
 		KeReleaseInStackQueuedSpinLock(&LockHandle);
 #else
-		KeReleaseSpinLock(&pMPDrvInfo->MPIOExtLock, SaveIrql);
+		KeReleaseSpinLock(&pwzvolDrvInfo->MPIOExtLock, SaveIrql);
```
lundman: Been tempted to remove the _AMD64_ here, as it appears to be just "more efficient", and the standard lock will work fine. It makes it harder to read, for practically no reason (only run when stopping).  Or is there some other reason to use `KeReleaseInStackQueuedSpinLock()` variants?

ZFSin/zfs/lib/libspl/posix.c
```diff
@@ -54,6 +55,13 @@ int fsync(int fd) {
 	return 0; 
 }
 
+const char* getexecname(void)
+{
+	__declspec(thread) static char execname[32767 + 1];
+	GetModuleFileNameA(NULL, execname, sizeof(execname));
+	return execname;
+}
+
```
lundman: Nice

ZFSin/zfs/include/sys/zfs_context.h
```diff
@@ -64,7 +64,22 @@
 
 #endif
 
-#define noinline
+#ifdef ENOTACTIVE
+#undef ENOTACTIVE
+#endif
+
+#define ENOTACTIVE EOWNERDEAD
```
lundman: I believe these should be moved to `spl/include//sys/errno.h` - I think I already have done in `rp_synth` branch, but not quite ready for commit, so just leave them be for now.

nazar554:
I think that it's still possible to attach from Visual Studio to kernel debugger while CMake project folder is open. The only problems are access to MSBuild-only stuff like the driver verifier, and the fact that `FindWDK.cmake` can't do ARM/ARM64 builds yet.

Also with CMake the userspace tools can be built with MinGW or Clang. (at least until LLVM/Clang starts supporting Windows drivers the kernel space is `cl.exe` only).

Using `.vcxproj` is fine too but needs refactoring. I think a good idea would be to use property pages (`.props`) for each library to make it cleaner (maybe in separate PR).

lundman:
I have merged into master some of the cleanup I have done. The biggest change is there is now a new `zfs_vnop_windows_mount.c` to keep that code separate. The rest is mostly moving helper functions from `zfs_vnop_windows.c` into `zfs_vnop_windows_lib.c`.
Hopefully it won't be too difficult to rebase.

nazar554:
Rebased. I see the new stuff requires Windows 10 APIs (QueryOnCreate related structures), will raise default `WDK_TARGET`.

Build is failing, maybe because I have higher WDK version (10.0.18362.0):
```
C2039	'Flags': is not a member of '_QUERY_ON_CREATE_ECP_CONTEXT'	 ZFSin\zfs\module\zfs\zfs_vnops_windows.c	1287
```

There is a comment in header:
```
BEWARE: This is an internal structure only and should not be used by filters as
//          it may change from release to release.  The proper way to use this
//          is by using the FltRequestFileInfoOnCreateCompletion
//          and FltRetrieveFileInfoOnCreateCompletion routines.
//          This structure changed between RS5 and 19H1
```

lundman:
Ah ok, lets just #ifdef `QUERY_ON_CREATE_ECP_CONTEXT` around the code

nazar554:
Done, also removed `PDB` stuff, doesn't seems it was necessary after all

lundman:
If you are happy with these changes you should probably squash it into one/few commits. ?

nazar554:
Sure will do today

nazar554:
Rebased

lundman:
So the new port is cmake only, and I see you have tests for ZFSIN_SIGNTOOL_CERTNAME being set,
what is the usual way this is done? I should presumably set ZFSIN_SIGNTOOL_CERTNAME to my certificate *somewhere*, which is a place not checked into git. Does cmake have a local build settings? Or just Win envvar set?

nazar554:
It is in the CMake cache, the certificate should be imported in Windows certstore. Setting are editable from Visual Studio [build configuration dropdown](https://docs.microsoft.com/en-us/cpp/build/customize-cmake-settings?view=msvc-160), or just using `-D` commandline option

lundman:
Ah so in CMakeSettings.json file? I assumed it is to be checked-in, when everyone building will want their own? guess it's not so important at this moment, only kernel builds so far.

nazar554:
I think that `CMakeSettings.json` is VS-specific. For CI purposes it's most likely setup using command-line.

Right now anyone who builds needs to `makecert` and import their own certificate and setup the variables. (also make it trusted, so driver can be used in Test signing mode)
If the certificate is inside the repo, it still needs to be imported to Windows Certificate Store somehow (maybe powershell or other command), probably during the configure step of CMake.
There is also an option to use `WDK Test cert, username, 1234567890` but I am not sure if it's always created when WDK is installed and it's name not same for everyone. Also not sure if it's trusted (for Test signing mode).

For "official" release builds certificate should be kept private, although the process is [now complicated](https://www.osr.com/blog/2020/10/15/microsoft-driver-updates-allowed-win7-win8/). I think before `code-signing` certificates from a few CAs were trusted for drivers, now it seems cross-signing is dead.

PR #168, opened 2019-10-04
Removing warnings

lundman:
Starting to chip away at the warnings.

* Added a list of warnings to ignore - generally the Windows ones that are not so useful

The "May truncate" warning is annoyingly verbose but can be useful, so left it in.


lundman:
OK, I believe I have removed the bulk of the noisy warnings, the remaining warnings I'll have to go through one-by-one as they are probably indicating something that should be fixed.

I'm not happy that adding ignore-warnings to the project seems to be ignored at times, so then I also had to add them to `zfs_context.h`. If that is the case, presumably it'd be better to only use the latter.

I've been increasingly unhappy using `#ifdef _WIN32` around `int fd;` and the Windows `HANDLE fd;` I think I'll float the idea to the openzfs devs about creating a `z_filetype fd;` and use that in those places. Mostly in import, and send/recv code.

We are forced to use HANDLEs, as the `open(int )` call can not open the disk devices. So perhaps then a shim layer that defines `open(HANDLE) read(HANDLE) write(HANDLE) close(HANDLE) lseek(HANDLE)` etc, would be the least changes of the shared codebase.
It is unattractive to replace the OS calls - like that of `open()` but I don't think upstream would be excited about `zopen()` etc everywhere.

lundman:
Actually, having thought about it for a bit. If we are going to implement own `open()` and companions, they might as well take `int fd` and internally handle the HANDLE conversion. Even if it is initially a straight type-cast and "hope it fits", it can eventually be extended to be done properly.
If something clever could be done to make sure that actual `int fd;`'s aren't passed in, that would be even better (DEBUG builds).

PR #173, opened 2019-10-09
Add appveyor artifact nightlies

TheCakeIsNaOH:
This does a couple of things-
1. Makes appveyor build the inno setup installer.
2. Publish the installer to appveyor artifacts, so making a nightly build
3. Changes the build from debug to release as that is needed for the installer.
4. Change from signtool arguments being in the script to being in the inno builder environment. This is needed so both appveyor and individual builders can build the installer without having to change the script.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,8 +1,15 @@
 version: 1.0.{build}
 image: Visual Studio 2017
-configuration: Debug
+configuration: Release
 platform: x64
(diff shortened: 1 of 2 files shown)


TheCakeIsNaOH:
Cool it actually works- https://ci.appveyor.com/project/lundman/zfsin/builds/27977348/artifacts

PR #174, opened 2019-10-10
Wosix

lundman:
Attempt to remove all the `#ifdef _WIN32` littering around the userland code base. The biggest changes is that we can not use the CRT calls that take `int fd;` - like `open()`. As they are not full citizens (can't open disk device nodes, nor handle the namespace lookup`.) The proper way is to use `HANDLE h = CreateFile()` to open things. 

We then wrap all the POSIX calls into our own version, and "hope" that HANDLE will fit in the smaller `int` type. Luckily, userland does not consume many file descriptors, so it has not yet been an issue. However, since the conversion use macros `HTOI()` and `ITOH()` it can be extended to handle the type conversion. (Presumable with a process style file descriptor array holding the real `HANDLE` types - like Unix already does). The challenge will be when userland passes the `int/HANDLE` to kernel.

Another option is to talk upstream into a wrapper type `typedef int z_fd;` to use everywhere we pass `int fd;` around, including `z_cmd_t` kernel passing.

There is a special case in the usage of `socketpair()` `ssread()` that still maintain a `#ifdef`.

`efi_rdwr.c` could still need cleanup.

### DIFF
--- a/ZFSin/zfs/lib/libzfs/libzfs_import.c
+++ b/ZFSin/zfs/lib/libzfs/libzfs_import.c
@@ -1093,7 +1093,7 @@ typedef struct rdsk_node {
 	char *rn_parent;
 #endif
 	int rn_num_labels;
-	HANDLE rn_dfd;
+	int rn_dfd;
 	libzfs_handle_t *rn_hdl;
 	nvlist_t *rn_config;
 	avl_tree_t *rn_avl;
@@ -1292,6 +1292,7 @@ void signal_alarm(int foo)
 }
 #endif
 
+#ifndef _WIN32
 static void
 zpool_open_func(void *arg)
 {
@@ -1452,7 +1453,9 @@ fprintf(stderr, "%s: enter\n", __func__); fflush(stderr);
 	rn->rn_config = config;
 	rn->rn_num_labels = num_labels;
 }
+#endif
 
+// Should probably rename this, and only define one.
 static void
 zpool_open_func_win(void *arg)
 {
@@ -1560,7 +1563,7 @@ zpool_open_func_win(void *arg)
 		* Try to read the disk label first so we don't have to
 		* open a bunch of minor nodes that can't have a zpool.
 		*/
-		check_slices(rn->rn_avl, fd, rn->rn_name);
+		check_slices(rn->rn_avl, HTOI(fd), rn->rn_name);
 	}
 
 	if ((zpool_read_label_win(fd, drive_len, &config, &num_labels)) != 0) {
@@ -1585,7 +1588,7 @@ zpool_open_func_win(void *arg)
  * Given a file descriptor, clear (zero) the label information.
  */
 int
-zpool_clear_label(HANDLE fd)
+zpool_clear_label(int fd)
 {
 	struct _stat64 statbuf;
 	int l;
@@ -1833,7 +1836,7 @@ zpool_find_import_impl(libzfs_handle_t *hdl, importargs_t *iarg)
 			slice->rn_name = zfs_strdup(hdl, name);
 			slice->rn_parent = zfs_strdup(hdl, path);
 			slice->rn_avl = &slice_cache;
-			slice->rn_dfd = dfd;
+			slice->rn_dfd = HTOI(dfd);
 			slice->rn_hdl = hdl;
 			slice->rn_nozpool = B_FALSE;
 			avl_add(&slice_cache, slice);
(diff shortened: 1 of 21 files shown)


lundman:
note-to-self: include new files.

ZFSin/zfs/lib/libzfs/libzfs_import.c
```diff
@@ -1560,7 +1563,7 @@ zpool_open_func_win(void *arg)
 		* Try to read the disk label first so we don't have to
 		* open a bunch of minor nodes that can't have a zpool.
 		*/
-		check_slices(rn->rn_avl, fd, rn->rn_name);
+		check_slices(rn->rn_avl, HTOI(fd), rn->rn_name);
```
lundman: Uses direct `HTOI()` macro as this is `zpool_open_func_win()` - ie the `_win()` version of the whole function.

ZFSin/zfs/lib/libzfs/libzfs_import.c
```diff
@@ -1833,7 +1836,7 @@ zpool_find_import_impl(libzfs_handle_t *hdl, importargs_t *iarg)
 			slice->rn_name = zfs_strdup(hdl, name);
 			slice->rn_parent = zfs_strdup(hdl, path);
 			slice->rn_avl = &slice_cache;
-			slice->rn_dfd = dfd;
+			slice->rn_dfd = HTOI(dfd);
```
lundman: This one isn't a `_win()` specific version, so it shouldn't use `HTOI()` - so it should probably be renamed to `_win()` in future (or rather, this will live in the `.../os/windows/` location in the future repository.

ZFSin/zfs/lib/libzfs/libzfs_pool.c
```diff
@@ -2886,7 +2886,7 @@ zpool_open_delay(int timeout, const char *path, int oflag)
 //		fd = open(path, oflag);
 //	}
 
-	return (fd);
+	return (HTOI(fd));
```
lundman: Whole function is Windows version.

ZFSin/zfs/lib/libzfs/libzfs_sendrecv.c
```diff
@@ -137,7 +137,7 @@ ssread(void *buf, size_t len, FILE *stream)
 {
 	size_t outlen;
 #ifdef WIN32
-	/* Windows we have a SOCKET type here*/
+	/* Windows we have a SOCKET type here - can't change fread() call. */
 	DWORD rv;
 	if (!ReadFile(stream, buf, len, &rv, NULL))
 		return 0;
```
lundman: Here is one of the remaining `#ifdef _WIN32`. Unix variant uses `fread()` as they are using actual `FILE *` - whereas we use `SOCKET`. Technically, it could call `read()` here, as that calls `wosix_read()` - but I think that would be more confusing. So it is more clear (I think) to call `ReadFile()` to highlight the special case.

ZFSin/zfs/lib/libzfs/libzfs_sendrecv.c
```diff
@@ -522,7 +503,11 @@ cksummer(void *arg)
 	umem_cache_destroy(ddt.ddecache);
 	free(ddt.dedup_hash_array);
 	free(buf);
-#ifndef WIN32
+#ifdef _WIN32
+	// We don't want to wrap fclose() so this is a special
+	// place we manually close the handle from socketpair().
+	CloseHandle(ofp);
+#else
 	(void) fclose(ofp);
```
lundman: And the matching special case for `SOCKET` close. We could call `close()` / `wosix_close()` here (I did make that special code to detect sockets, why did I do that then not use it). But it seems a bit unclear to mix `fclose()` and `close()`.

ZFSin/zfs/lib/libzfs/libzfs_sendrecv.c
```diff
@@ -1301,7 +1286,7 @@ send_progress_thread(void *arg)
 		0, NULL, OPEN_EXISTING, 0, NULL);
 	libzfs_handle_t tmp;
 	if (h != INVALID_HANDLE_VALUE) {
-		tmp.libzfs_fd = h;
+		tmp.libzfs_fd = HTOI(h);
```
lundman: Another `#ifdef _WIN32` case. The progress_thread must re-open the `/dev/zfs` device node to be allowed to issue `ioctl()s` on it. (Or it will block to wait for `ioctl(send);` to finish, rendering 'progress' useless).
This could be changed to `libzfs_open()` if that is considered cleaner.

PR #178, opened 2019-10-11
Fastio

lundman:


lundman:
Something goes wrong here with writes; if we start a simple:
```
dd if=/dev/random if=/e/file.txt bs=128k
```

It will initially write as expected, but eventually we end up in a peculiar state.

The writing thread appears to do:
```
fastio_write()
{
ret = FsRtlCopyWrite(FileObject, FileOffset, Length, Wait, LockKey, Buffer, IoStatus,
}
```

Where `FsRtlCopyWrite()` returns false all the time ( it appears full ).

We do get calls to `fs_write()`, which end up taking the path
```
			if (!CcCopyWriteEx(fileObject,
```

Ie, this write should go to cache, as `IRP_PAGING_IO` is not set.

I totally also expect to see calls to `fs_write(IRP_PAGING_IO)` which actually calls `zfs_write()` to commit the data to disk, but this is not happening. Why?

So it would appear we get fastio writes until memory is exhausted and it stalled. Meanwhile, we do not get the lazywrites, they appear to not happen. We do not seem to have any threads stuck waiting on locks, nor are the lazy/fastio lock calls even called.

PR #179, opened 2019-10-12
Readme update with nightly builds information and signed builds instructions

TheCakeIsNaOH:


README.md
```diff
@@ -251,32 +251,28 @@ not set, it will mount "/pool" as "C:/pool".
...
-corner of the screen.
 
 After that either 
 
-* Run OpenZFSOnWindows.exe installer to install.
-* Windows should popup the "Windows can't verify the publisher of this driver software"
-* Click "Install this driver software anyway"
+* Run OpenZFSOnWindows.exe installer to install
+* *Would you like to install device software?* should pop up, click install
+  * If installing an unsigned release, click "Install anyway" in the "unknown developer" popup
```
lundman: Running the installer, when I produce it, should not have "unknown developer" warning any more?

README.md
```diff
@@ -412,5 +408,7 @@ Also, there is [`kstat`](https://openzfsonosx.org/wiki/Windows_kstat) to dynamic
 
 # Nightly builds
 
-We do not have nightly builds.  
-However, from time to time, we do quick builds for people to test a fix. [Get it here](https://openzfsonosx.org/wiki/Windows_builds).
+There are nightly builds available at [AppVeyor](https://ci.appveyor.com/project/lundman/zfsin/branch/master/artifacts)  
+- These builds are currently not signed and therefore require test mode to be enabled.
+
+There also are some historical test builds [available here](https://openzfsonosx.org/wiki/Windows_builds).
```
lundman: I will most likely keep producing signed builds for the testers to try out - unless there is a way to do the automatic build with certificate.

README.md
TheCakeIsNaOH: Correct, the signed releases should be good to go without test mode.

A fresh install of Windows 10 1607 or higher with secure boot enabled, is the only exception, since that requires drivers to be signed with a EV cert to install them.

README.md
TheCakeIsNaOH: Fix pushed.

PR #182, opened 2019-10-19
Appveyor signing

TheCakeIsNaOH:
So this should allow signing of the appveyor binaries. 

There are a couple things to do to set it up on your end, but it should not be too hard.

I will come back and add more info on how to set this up.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,7 @@ environment:
   cat_location: 'C:\projects\zfsin\x64\Release\ZFSin\zfsin.cat'
   sys_location: 'C:\projects\zfsin\x64\Release\ZFSin\zfsin.sys'
   cross_cert: 'C:\projects\zfsin\PRIVATE-Signing\crosscert.crt'
+  pfx_path: 'C:\projects\zfsin\PRIVATE-Signing\ZFSin-private-cert.pfx'
   signtoola: '/Ssigntoola=C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x64\signtool.exe sign /a /t http://timestamp.digicert.com /fd sha1 /d $qOpenZFS on Windows$q $f'
   signtoolb: '/Ssigntoolb=C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x64\signtool.exe sign /a /as /tr http://timestamp.digicert.com /td sha256 /fd sha256 /d $qOpenZFS on Windows$q $f'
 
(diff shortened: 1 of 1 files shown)


TheCakeIsNaOH:
There a couple of things to do to set this up-
1. Create an SSH key pair `ssh-keygen -t rsa` and add the public key to github. 
2. Take the ssh private key and encrypt it here- https://ci.appveyor.com/tools/encrypt, then take the encrypted value and use it to replace the priv_key secure variable that is in the PR. 
3. Create a private repository called `PRIVATE-Signing` (can be called something else but will need to change paths in yaml file.
4. Add CA cross signing cert and also your signing cert with the private key(in `.pfx`) and add them to the private-signing repository.
5. Encrypt the password to the `.pfx` the same as above with the ssh and change the `pfx_key secure` variable.
6. If needed, change the `cross_cert` and `pfx_path` variables to point to the correct file names.
7. Add a git submodule(cloning with SSH), also make sure it tracks the master of the submodule repository. ie `git submodule add -b master [ssh url to Git repo]`

There is a reference for some of this here-
https://www.appveyor.com/docs/how-to/private-git-sub-modules/

TheCakeIsNaOH:
Right now, it is set up so both PRs and main branches will build.

On a PR, it will test sign, and with main branches, it will sign with a real cert. That is what the IF statements are for, so it will skip setting up the submodule and signing with the real cert if the build is for a PR.

The secure variables are not decrypted on a PR, so even if the if statements where removed in a malicious PR, the build would just fail as the credentials to access the real cert would not be accessible.

TheCakeIsNaOH:
There are still things that can be improved, such as the signtool arguments in the after build script, and figuring out why the PowerShell scripts demanded double spacing to work. However, it is good enough for a first attempt.

lundman:
Does #3 need to be a private repository? It would mean changing from free to paid github account, and I would need to budget the $25/month charge it adds :)

TheCakeIsNaOH:
Github has private repositories for free now.
https://github.blog/2019-01-07-new-year-new-github/

Or you could have it hosted at Gitlab, if you wanted, since the access is via ssh which Gitlab supports.

Alternatively, if you could to host a private git repo yourself, but that would be a lot more work.

TheCakeIsNaOH:
It does need to be private, otherwise, people could download your private certificate, which while still having a password, is not secure at all.

lundman:
Alas, this is a Team Project - who knows why - so if I click Private repository, GH goes to the Upgrade page. I could change to be the only contributor, and it is only $9/month. But at the moment, I don't think signing the compiled PR to be adding much - that it compiles and tests is very useful, but only devs will care about PR builds, regular users shouldn't be installing those builds.

Any opinions everyone?

TheCakeIsNaOH:
>But at the moment, I don't think signing the compiled PR to be adding much - that it compiles and tests is very useful, but only devs will care about PR builds, regular users shouldn't be installing those builds.

This does not sign PR builds, it only signs non-PR builds. If the build is a PR, it signs with the test certificate which does nothing, while if the build is not a PR, then it builds and signs with your real cert.

TheCakeIsNaOH:
Closing since it seems like there is no interest/need for this. 

Feel free to use this code or ping me if later signed appveyor builds are wanted.

PR #199, opened 2019-12-20
ZFS Windows related stability fixes

VasilMalinov:
Problem 1: Driver verifier breaks when ZFS tries to allocate Pool with the tag NonPagedPool                
Reason: Microsoft recommends to use NonPagedPoolNx after Windows 8 for all memory that does not need execution privilege.
Solution: Some of the places in driver were changed to request allocations from the NonPagedPoolNx.

Problem 2: Driver verifier breaks with memory access violation after calling the CreateReparsePoint() function. It causes a crash when Driver Verifier is not active. 
Reason - The LPCWSTR could be non-NULL terminated when passed in and causes the buffer to be overran.
Solution: Change LPCWSTR to PCUNICODE_STRING since there is no need for the function to recalculate the size of the buffer already built by the caller in Unicode string format.

Problem 3: Driver verifier breaks when ZFS’s dispatcher function is called at DISPATCH_LEVEL.
Reason: StorPort can call in the dispatcher at IRQL DISPACH_LEVEL but FsRtlEnterFileSystem() can only be called at IRQL<= APC_LEVEL. 
Solution: Check the IRQL level and do not call FsRtlEnterFileSystem() if it is greater than APC_LEVEL.

Problem 4: In the dispatcher() routine when the IRP is returned from a processing subroutine we sometimes get a BSOD.
Reason: The Irp pointer is accessed after returning from the dispatcher’s processing routine. When the IRP is returned with STATUS_PENDING and is already completed asynchronously (and freed) we get the memory access violation.
Solution: Do not use the Irp pointer if STATUS_PENDING is returned by the dispatcher’s subroutine.

### DIFF
--- a/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
+++ b/ZFSin/zfs/module/zfs/zfs_vnops_windows.c
@@ -4592,7 +4592,6 @@ fsDispatcher(
 		break;
 	}
 
-
 	/* Re-check (since MJ_CREATE/vnop_lookup might have set it) vp here, to see if
 	 * we should call setsize 
 	 */
@@ -4645,8 +4644,9 @@ dispatcher(
 )
 {
 	BOOLEAN TopLevel = FALSE;
+	BOOLEAN AtIrqlPassiveLevel;
 	PIO_STACK_LOCATION IrpSp;
-	NTSTATUS Status;
+	NTSTATUS Status = STATUS_NOT_IMPLEMENTED;
 	uint64_t validity_check;
 
 	// Storport can call itself (and hence, ourselves) so this isn't always true.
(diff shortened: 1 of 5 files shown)


ZFSin/zfs/module/zfs/zfs_vnops_windows.c
```diff
@@ -4666,76 +4666,82 @@ dispatcher(
...
 			Irp->IoStatus.Information, major2str(IrpSp->MajorFunction, IrpSp->MinorFunction));
 	}
 
 	// Complete the request if it isn't pending (ie, we called zfsdev_async())
-	ASSERT(validity_check == *((uint64_t *)Irp));
-
 	if (Status != STATUS_PENDING)
-		IoCompleteRequest(Irp, Status == STATUS_SUCCESS ? IO_DISK_INCREMENT : IO_NO_INCREMENT);
+	{
+		if (validity_check == *((uint64_t *)Irp)) {
```
lundman: Actually, the `validity_check` was just something I added when the IRPs were being corrupted (something I was doing wrong). I believe I since then fixed that, I've not had any for a while. But anyway, certainly never meant for it to be part of the code. But then you are using it like ASSERT anyway. But at some point, the `validity_check` code should all be deleted. Let's leave it for now, and remove it when we are sure it doesn't trigger :)

VasilMalinov: I will take this description in my future changes. 
One of the reason to crash the IRP is: When the IRP timeout expired and system completed IRP behind our workflow.  As result we access invalid memory area.

Issue #201, opened 2020-01-18
PFN_LIST_CORRUPT (4e)


lundman:
As reported by Datacore, on version 0.21.

I created a pool with dedup=on and carved out a zvol using these commands.

```
zpool create -O dedup=on -O casesensitivity=insensitive -O compression=lz4 -O atime=off -o ashift=12 zdedup PHYSICALDRIVEx
zfs create -s -V 3GB -b 512 zdedup/Z._629e5981-fa3f-11e9-84be-00155d050e1d_-00000060
```

I run I/O to the presented windows disk and after a short while, after I/Os have been stopped even, I get this BSOD.  I took a quick look at it and it seems that the slab is being reused while it is disposed of.

<details>
  <summary> code diff </summary>
  <p>

```
*******************************************************************************
*                                                                             *
*                        Bugcheck Analysis                                    *
*                                                                             *
*******************************************************************************

PFN_LIST_CORRUPT (4e)
Typically caused by drivers passing bad memory descriptor lists (ie: calling
MmUnlockPages twice with the same list, etc).  If a kernel debugger is
available get the stack trace.
Arguments:
Arg1: 000000000000009a,
Arg2: 0000000000084fab
Arg3: 0000000000000006
Arg4: 0000000000000002

Debugging Details:
------------------


DUMP_CLASS: 1

DUMP_QUALIFIER: 401

BUILD_VERSION_STRING:  14393.2879.amd64fre.rs1_release_inmarket.190313-1855

SYSTEM_MANUFACTURER:  Microsoft Corporation

VIRTUAL_MACHINE:  HyperV

SYSTEM_PRODUCT_NAME:  Virtual Machine

SYSTEM_SKU:  None

SYSTEM_VERSION:  Hyper-V UEFI Release v4.0

BIOS_VENDOR:  Microsoft Corporation

BIOS_VERSION:  Hyper-V UEFI Release v4.0

BIOS_DATE:  01/30/2019

BASEBOARD_MANUFACTURER:  Microsoft Corporation

BASEBOARD_PRODUCT:  Virtual Machine

BASEBOARD_VERSION:  Hyper-V UEFI Release v4.0

DUMP_TYPE:  1

BUGCHECK_P1: 9a

BUGCHECK_P2: 84fab

BUGCHECK_P3: 6

BUGCHECK_P4: 2

BUGCHECK_STR:  0x4E_9a

CPU_COUNT: 6

CPU_MHZ: da4

[...]

PR #227, opened 2020-03-23
Replacing all the instances of IoFreeMdl with UnlockAndFreeMDL routine

vrajendra-datacore:
Problem: Incorrectly releasing the MDL chain was leading to memory leaks.

Cause: MDLs' first node was being released instead of all the nodes in MDL chain. Thereby leading to memory leaks.

Fix: Freeing up all the nodes in the MDL chain after unlocking each physical pages described by the MDL. We have a common routine UnlockAndFreeMDL being called wherever we need to free up the MDL list.

### DIFF
--- a/ZFSin/spl/module/spl/spl-windows.c
+++ b/ZFSin/spl/module/spl/spl-windows.c
@@ -199,6 +199,24 @@ fnv_32a_buf(void *buf, size_t len, uint32_t hval)
 	return hval;
 }
 
+/*
+ * Function to free a MDL chain
+ */
+void UnlockAndFreeMDL(PMDL Mdl)
+{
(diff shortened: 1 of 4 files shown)


imtiazdc:
I have reasons to believe that the issue [PFN_LIST_CORRUPT (4e)](https://github.com/openzfsonwindows/ZFSin/issues/201) too must have been fixed by this change.

Before this change, MDLs (in 5 places) were not being unlocked and freed in a loop, instead only the head of the MDL was being unlocked and freed.

I believe this was causing the MM to reuse the remaining pages in other context without dropping/clearing the reference counting.

However, when the consumer of those remaining pages from the other context tried to free, the system bug checked saying it was already locked for IO (based on the non-zero reference count).

PR #228, opened 2020-03-27
Go beyond 16 zvols presented to disk manager

datacore-tlaurent:
Opened up the zvol LUN presentation through StorPort to be able to announce up to 32640 zvols.

- By default the maximum amount of ZVOL LUNs is set to 400 (from 16 today) and customization through kstat will be added through a future change.

- Also by default the distribution to the targets is set to 32 LUNs per target.  This number will limit the affected LUNs when a SCSI target reset is issued (kstat customizable in the future).

- The SCSI Report LUNs has been fixed in order to return a list of more than 1 LUN.

- The MPIO support has been fixed to take into consideration the bCombineVirtDisks parameter and by default it is (and remains) FALSE.  This means that, no matter the amount of HBA extensions that are called by StorPort through FindAdapter, only 1 adapter will announce LUNs and service SCSI requests, and only 1 path will show up in MPIO.
When set to TRUE all adapters defined will respond to SCSI Inquiry and Report Luns, and will therefore enable those LUNs for multipathing in MPIO.
At that point the Zvol disk will show as many adapter paths as defined, set by default to failover-only with only 1 path active/optimized and others in Standby.  So when setting this parameter to TRUE users should first define the "OpenZFS WinZVOL" VendorId/ProductId in the MPIO console so they can be grabbed and bundled by MPIO.
Because ZFSin's StorPort interface is a virtual miniport, MPIO support is somewhat moot so this parameter should remain FALSE.

- The VendorSpecific field in the SCSI Inquiry command is now filled with the pathid/targetid/lunid used to present that ZVOL, unless a more valuable 20 bytes value is thought of. This is informational only and does not have functional significance.

- Removed the potentially costly call to StorPortGetLogicalUnit() and fetch the zv_state structure instead since everything we need is there and the LU extension is not necessary (kept for now).

- Changed the MapBuffers option to map all SRB's data buffers into system address space and take out the StorPortGetSystemAddress() call for I/Os.  We do not have any physical hardware to DMA to so ScatterGather lists are of no use for ZFSin.  Moreover in some instances, buffers were not mapped to system address space and could lead to BSOD so this takes care of those too.

ref: ssv-18396

### DIFF
--- a/ZFSin/zfs/module/zfs/zfs_windows_zvol_scsi.c
+++ b/ZFSin/zfs/module/zfs/zfs_windows_zvol_scsi.c
@@ -597,13 +597,6 @@ ScsiOpModeSense(
 /**************************************************************************************************/     
 /*                                                                                                */     
 /**************************************************************************************************/     
-VOID ReverseIndian(UCHAR *x)
-{
-	UCHAR t;
-	UCHAR* d = (UCHAR*)x;
-	t = d[0]; d[0] = d[3]; d[3] = t; t = d[1]; d[1] = d[2]; d[2] = t;
-}
-
 UCHAR
 ScsiOpReportLuns(                                     
                  __in __out pHW_HBA_EXT         pHBAExt,   // Adapter device-object extension from StorPort.
@@ -637,8 +630,7 @@ ScsiOpReportLuns(
         }
     } // else: we chose to not report any LUN through that HBA (see FindAdapter routine).
 
-	*((ULONG*)&pLunList->LunListLength) = totalLun * sizeof(pLunList->Lun[0]);
-	ReverseIndian(pLunList->LunListLength);
+	*((ULONG*)&pLunList->LunListLength) = RtlUlongByteSwap(totalLun * sizeof(pLunList->Lun[0]));
 	pSrb->DataTransferLength = FIELD_OFFSET(LUN_LIST, Lun) + (GoodLunIdx * sizeof(pLunList->Lun[0]));
 
     return status;
--- a/ZFSin/zfs/module/zfs/zfs_windows_zvol_scsi.c
+++ b/ZFSin/zfs/module/zfs/zfs_windows_zvol_scsi.c
@@ -655,7 +655,6 @@ wzvol_WkRtn(__in PVOID pWkParms)                          // Parm list pointer.
 	ULONG                     startingSector,
 		sectorOffset,
 		lclStatus;
-	PVOID                     pX = NULL;
 	UCHAR                     status;
 
 	zvol_state_t *zv;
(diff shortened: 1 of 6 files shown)


ZFSin/zfs/module/zfs/zfs_windows_zvol_scsi.c
```diff
@@ -639,55 +597,49 @@ ScsiOpModeSense(
 /**************************************************************************************************/     
 /*                                                                                                */     
 /**************************************************************************************************/     
+VOID ReverseIndian(UCHAR *x)
+{
+	UCHAR t;
+	UCHAR* d = (UCHAR*)x;
+	t = d[0]; d[0] = d[3]; d[3] = t; t = d[1]; d[1] = d[2]; d[2] = t;
+}
+
```
lundman: Is this supposed to be `ReverseEndian()` ? As in changing to BigEndian (presumably scsi operates on big)? Should we not use `htonl()` in that case?

datacore-tlaurent: Yes, htonl() works but after checking the WDK I found that they recommend using the Rtl version of the WDK API: https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlulongbyteswap
So I used that, retested and committed. Thanks.

lundman: Yeah it seems on windows htonl and family is in winsock2, so it is a whole thing, even if we are linking against it in this project. Thanks for the update.

lundman:
Ah hmm I merged the first PR and now this one needs a small conflict fix - my apologies.

ZFSin/zfs/module/zfs/zfs_windows_zvol_scsi.c
```diff
@@ -642,52 +600,38 @@ ScsiOpModeSense(
...
+				if (pSrb->DataTransferLength >= FIELD_OFFSET(LUN_LIST, Lun) + (GoodLunIdx * sizeof(pLunList->Lun[0])) + sizeof(pLunList->Lun[0])) {
+					pLunList->Lun[GoodLunIdx][1] = (UCHAR)i;
+					GoodLunIdx++;
+				}
+			}
         }
-    }
+    } // else: we chose to not report any LUN through that HBA (see FindAdapter routine).
+
+	*((ULONG*)&pLunList->LunListLength) = RtlUlongByteSwap(totalLun * sizeof(pLunList->Lun[0]));
```
lundman: Actually, not sure why this looks so complicated, a simple
`pLunList->LunListLength = RtlUlongByteSwap(totalLun * sizeof(pLunList->Lun[0]));`
should work? Is LunListLength not ULONG ?

lundman: ` UCHAR LunListLength[4];`
Ah! and ew! :)  Ok, that makes sense.

datacore-tlaurent:
I resolved the conflict.  Ready for merge now.

Issue #109, opened 2019-03-01
What about dem icons?


lundman:
At some point we also need icons for the .exes and installer. That sort of thing is definitely not my wheelhouse so if someone happened to want to do something, that'd be great.

TheCakeIsNaOH:
@lundman Can this be closed now that there are icons?

lundman:
Cheers.

PR #234, opened 2020-04-02
(GH-109) Add icons to executables. 

TheCakeIsNaOH:
This adds icons to various executables. It uses the icon from OpenZFS, although this could be changed fairly easily in the future. 

From #109


lundman:
Neat, the Installer icon works and looks good.  The diff makes me think the executables should have icons, like `zpool.exe` `zfs.exe` but they do not seem to work.

Also get error when starting installer, it can be ignored, but probably indicates a problem:

Screen Shot 2020-04-03 at 9 35 07 (2)

TheCakeIsNaOH:
The icons on the individual `.exes` are working for me. They get added in the visual studio build, not by the inno compiler. They are working in the appveyor build, if you extract the installer the icons are there- 

https://ci.appveyor.com/api/buildjobs/0oqeuj0qorg2twf4/artifacts/OpenZFSOnWindows-release.exe
image

As for the warning when running the installer, I am not sure. What Inno setup version are you compiling with? You also could try running the installer from the appveyor build to see if it also shows the warning.

lundman:
Yep, total pebkac on the exe icons.

The error at start goes away if I shrink the openzfs-small.bmp to 25% (30% is NG). But could be something with the inno version I have, I will try to update first.

lundman:
Updating Inno seemed to have worked. Cheers.

PR #249, opened 2020-05-12
Fix zpool create returning -1 on successful creation of zpool

vrajendra-datacore:
Problem: The exit status code of zpool create command after successful completion returns -1

Cause: The function call sa_zfs_process_share inside zfs_share_proto function returns -1.
In function zfs_share_proto, we have for loop "for (curr_proto = proto; *curr_proto != PROTO_END; curr_proto++)", where in we iterate PROTO_END (3) times;  One of the arguments to sa_zfs_process_share is proto_table which is defined with only 2 entries (one for nfs, another for smb); (Third entry is supposedly reserved for afp but not explicitly defined in proto_table)
In the third iteration, proto_table[2] is invalid and hence returns with failure.

Fix: Since AFP protocol is available for Mac platform, we keep the code related to AFP protocol out of the build process by using #ifdef __APPLE__.


lundman:
As it happens, AFP has been deprecated by APPLE, and I have taken it out of the new OSX port, so it will vanish here as well at some point.  So this PR can be merged as-is, and I'll do so in a day or so, but if you wanted to change it, feel free to delete the AFP relevant sections instead.

PR #253, opened 2020-05-19
Eliminate memory allocation in zvol queueing from StorPort I/O.

datacore-tlaurent:
Use the WorkRtnParms allocation to include space for the workitem, thus eliminating an extra allocation. Used the SRB extension to include the WrkRtnParms block. no more extra dynamic allocation necessary for each zvol read/write I/O that gets queued in a work item queue.


datacore-tlaurent:
In parallel I am working on eliminating the mem allocation that is used to build the I/O descriptor that is queued in a work item queue.  We can use the SRB extension for that.  Just want to do incremental changes.

PR #262, opened 2020-06-17
Fixed race condition in zfs destroy

datacore-tlaurent:
Problem: random BSODs happen when the zv (zvol_state_t*) control block happens to be freed while outstanding I/Os are still in the StorPort acceptance pipeline.  Once I/Os make it in the zil log the close routine already makes sure to flush the zil before freeing the zv but some I/Os could still be in the path length's front end portion.

Fix: Front-end I/Os - entering through StorPort - must be be all drained before the zfs destroy command deletes the zv control block.  

Also clearing the target/lun id *before* invoking the close shuts off front-end I/Os earlier and allows for a speedier close.  
Note: Comments explaining the reference count/remove lock mechanism are embedded with code (see ref: SSV-18807 in the code).

### DIFF
--- a/ZFSin/zfs/module/zfs/zfs_windows_zvol_scsi.c
+++ b/ZFSin/zfs/module/zfs/zfs_windows_zvol_scsi.c
@@ -65,45 +65,162 @@
 /*
  * We have a list of ZVOLs, and we receive incoming (Target, Lun) requests that needs to be mapped
  * to the correct "zv" ptr.
+ *
+ * ssv-18807: fixed the race condition in the zvol destroy processing by adding remove lock logic
+ * to ensure no new I/O can be processed from the front end (StorPort) and all outstanding host I/Os 
+ * have left the pipeline.
+ *
+ * The zv control block starts to get protected in wzvol_assign_targetid() and this until wzvol_clear_targetid()
+ * is called.
+ * 
+ * Once wzvol_find_target(t,l) returns a valid pointer to the zv, that zv is protected via an extra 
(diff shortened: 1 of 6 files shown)


ZFSin/zfs/module/zfs/zfs_windows_zvol_scsi.c
```diff
@@ -65,45 +65,162 @@
 /*
  * We have a list of ZVOLs, and we receive incoming (Target, Lun) requests that needs to be mapped
  * to the correct "zv" ptr.
+ *
+ * ssv-18807: fixed the race condition in the zvol destroy processing by adding remove lock logic
+ * to ensure no new I/O can be processed from the front end (StorPort) and all outstanding host I/Os 
+ * have left the pipeline.
+ *
+ * The zv control block starts to get protected in wzvol_assign_targetid() and this until wzvol_clear_targetid()
+ * is called.
```
lundman, line 74: "and this until" ?

datacore-tlaurent: Upon return from the call to wzvol_clear_targetid() we are then 100% sure that no new front-end I/O can enter the processing pipeline.  Which makes it safe to then free the zv based on existing protections from the back-end processing logic.

lundman:
It's interesting to see other solutions to a problem, this uses Windows tools to accomplish it. I know ZFS primitives better, so I would probably have reached for a rwlock. Neat.

PR #267, opened 2020-06-22
Enabling WPP tracing in ZFSin

vrajendra-datacore:
WPP is enabled only on Release build.

dprintf call, which is predominantly used to log message is used to generate the trace logs.

There are also few compilation error fix like variable length format specifier like "%.*s" and "%.*S"

### DIFF
--- a/ZFSin/Trace.h
+++ b/ZFSin/Trace.h
@@ -1,0 +1,33 @@
+#pragma once
+
+#define WPPNAME		ZFSinTraceGuid
+#define WPPGUID		c20c603c,afd4,467d,bf76,c0a4c10553df
+
+#define WPP_CONTROL_GUIDS \
+    WPP_DEFINE_CONTROL_GUID(WPPNAME,(WPPGUID), \
+        WPP_DEFINE_BIT(MYDRIVER_ALL_INFO)             /* bit  0 = 0x00000001 */ \
+        WPP_DEFINE_BIT(TRACE_DRIVER)           /* bit  1 = 0x00000002 */ \
+        WPP_DEFINE_BIT(TRACE_DEVICE)              /* bit  2 = 0x00000004 */ \
+        WPP_DEFINE_BIT(TRACE_QUEUE)           /* bit  3 = 0x00000008 */ )
+
+#ifndef WPP_CHECK_INIT
+#define WPP_CHECK_INIT
+#endif
+
+
+#define WPP_FLAGS_LEVEL_LOGGER(Flags, level)                                  \
+    WPP_LEVEL_LOGGER(Flags)
+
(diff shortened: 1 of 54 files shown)


imtiazdc:
Thanks @vrajendra-datacore for the PR. 

@lundman this change has been tested in our environment. Let us know if you have any comments or concerns with this change.

On top of it, we plan to make changes to enable tracing using fine-grained levels. With that we can collect/see traces of our interests at the runtime. We also plan to introduce timestamps in the traces.

lundman:
It appears closer to a trace facility than a log printing one, akin to dtrace.

So keeping in mind that this might need to live upstream, what are `.tmh` files? Are they automatically generated when compiling? If so, they can live in the include directory, if not, they should probably be in our own includes.

Can the `#if RUN_WPP` which is in all source files be included via zfs_context_kernel.h or similar? Using _ _FILE__? (SPL files are 'ours' so that is ok, but the module/zfs/ files are generally shared).

nconnolly1:
WPP files are transient files created by the WPP pre-processor as part of a build.  They will change with every modification to the associated source file, so shouldn't be stored in the repo.  Ideally, the Visual Studio project should be set up so that the .tmh files are created in the intermediate directory rather than in the source directory.

I wish there was a good way of automating the #include of the .tmh files - unfortunately to handle this in an include file, you really want something like __PARENT_FILE__ to determine the name of the file including the tracing header (so yield the .tmh file name).  There's no standard way (that I've found) of getting this.

The best that I've come up with in the past is to use:

#define WPPFILE "<put base file name here>.tmh"
#include "Trace.h"

That way the #ifdefs can be hidden inside the trace.h header and the header can do nothing if either WPPFILE or RUN_WPP aren't defined.  The .tmh can be included with #include WPPFILE.  It also means there's a single place to change the #ifdefs if they ever need to be altered.

imtiazdc:
> Can the `#if RUN_WPP` which is in all source files be included via zfs_context_kernel.h or similar? Using _ _FILE__? (SPL files are 'ours' so that is ok, but the module/zfs/ files are generally shared).

@lundman zfs_context_kernel.h seems to be a common header file in shared code - module/zfs. Would you know if there is a common header file for SPL too?

lundman:
I was more concerned about the shared sources files in module/zfs/ like dmu.c, dsl_scan.c etc. At worst, the WPPFILE could be set with -D for each of those files individually, if upstream would rather not see that.  For "our" files, all of SPL and anything with *windows* in it, it is fine as is.

vrajendra-datacore:
@lundman We are working on raising another PR request devoid of any changes in shared common files

lundman:
Awesome - I would be OK to approve this as-is, but we'll need to redo things in common files when we merge with upstream. So if we can find a balance now, that is even better.

PR #268, opened 2020-06-23
Make vdev_disk_io_start & vdev_file_io_start non-blocking

aviksildc:
Problem: At present Read/Write IOs block in vdev_disk_io_start and vdev_file_io_start for completion. They synchronously wait for IOs to complete which reduces performance, especially on high latency disks.

Fix: This patch modifies the IO completion routine to do the post-processing there itself, instead of signaling an event. Since there are some restrictions in performing certain operations (blocking calls, mutex etc) in IO completion routine as it might run in higher IRQ level, the patch addresses this by queuing a new work item which does the post-processing, if necessary.

### DIFF
--- a/ZFSin/zfs/module/zfs/vdev_disk.c
+++ b/ZFSin/zfs/module/zfs/vdev_disk.c
@@ -450,10 +450,10 @@ vdev_disk_io_start_done(void *param)
 	}
 
 	UnlockAndFreeMdl(vb->irp->MdlAddress);
-	zio_delay_interrupt(zio);
 	IoFreeIrp(vb->irp);
 	kmem_free(vb, sizeof(vd_callback_t) + IoSizeofWorkItem());
 	vb = NULL;
+	zio_delay_interrupt(zio);
 }
 
 static VOID
@@ -619,12 +619,6 @@ vdev_disk_io_start(zio_t *zio)
 
 	/* Preallocate space for IoWorkItem, required for vdev_disk_io_start_done callback */
 	vd_callback_t *vb = (vd_callback_t *)kmem_alloc(sizeof(vd_callback_t) + IoSizeofWorkItem(), KM_SLEEP);
-
-	if (!vb) {
-		zio->io_error = EIO;
-		zio_interrupt(zio);
-		return;
-	}
 
 	vb->zio = zio;
 
--- a/ZFSin/zfs/module/zfs/vdev_disk.c
+++ b/ZFSin/zfs/module/zfs/vdev_disk.c
@@ -352,7 +352,6 @@ vdev_disk_close(vdev_t *vd)
 	if (vd->vdev_reopening || dvd == NULL)
 		return;
 
-
 	vd->vdev_delayed_close = B_FALSE;
 	/*
 	 * If we closed the LDI handle due to an offline notify from LDI,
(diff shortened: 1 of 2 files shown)


ZFSin/zfs/module/zfs/vdev_disk.c
```diff
@@ -439,6 +419,85 @@ vdev_disk_ioctl_done(void *zio_arg, int error)
...
+		abd_return_buf_copy_off(zio->io_abd, vb->b_addr,
+			0, zio->io_size, zio->io_abd->abd_size);
+	} else {
+		VERIFY3S(zio->io_abd->abd_size, >= , zio->io_size);
+		abd_return_buf_off(zio->io_abd, vb->b_addr,
+			0, zio->io_size, zio->io_abd->abd_size);
+	}
+
+	UnlockAndFreeMdl(vb->irp->MdlAddress);
+	zio_delay_interrupt(zio);
```
lundman, line 453: Can we make the `zio_delay_interrupt()` call be the last in the function.

ZFSin/zfs/module/zfs/vdev_disk.c
```diff
@@ -439,6 +419,85 @@ vdev_disk_ioctl_done(void *zio_arg, int error)
...
+
+	/* If IRQL is below DIPATCH_LEVEL then there is no issue in calling
+	 * vdev_disk_io_start_done() directly; otherwise queue a new Work Item
+	*/
+	if (KeGetCurrentIrql() < DISPATCH_LEVEL)
+		vdev_disk_io_start_done(vb);
+	else {
+		IoInitializeWorkItem(dvd->vd_DeviceObject, (PIO_WORKITEM)vb->work_item);
+		IoQueueWorkItem((PIO_WORKITEM)vb->work_item, DiskIoWkRtn, DelayedWorkQueue, vb);
+	}
```
lundman, line 497: Ah is that how you are supposed to do that. Cheers

ZFSin/zfs/module/zfs/vdev_disk.c
```diff
@@ -549,61 +608,60 @@ vdev_disk_io_start(zio_t *zio)
...
 	IO_STATUS_BLOCK IoStatusBlock = { 0 };
 	LARGE_INTEGER offset;
-	KEVENT Event;
 
 	offset.QuadPart = zio->io_offset + vd->vdev_win_offset;
 
+	/* Preallocate space for IoWorkItem, required for vdev_disk_io_start_done callback */
+	vd_callback_t *vb = (vd_callback_t *)kmem_alloc(sizeof(vd_callback_t) + IoSizeofWorkItem(), KM_SLEEP);
+
+	if (!vb) {
```
lundman, line 623: No need, if you call `kmem_alloc( ..., KM_SLEEP)` it will never return NULL, but sleep until there is memory.

ZFSin/zfs/module/zfs/vdev_file.c
```diff
@@ -328,13 +328,74 @@ vdev_file_close(vdev_t *vd)
...
+		abd_return_buf_copy_off(zio->io_abd, vb->b_data,
+			0, zio->io_size, zio->io_abd->abd_size);
+	} else {
+		VERIFY3S(zio->io_abd->abd_size, >= , zio->io_size);
+		abd_return_buf_off(zio->io_abd, vb->b_data,
+			0, zio->io_size, zio->io_abd->abd_size);
+	}
+
+	UnlockAndFreeMdl(vb->irp->MdlAddress);
+	zio_delay_interrupt(zio);
```
lundman, line 362: ditto

ZFSin/zfs/module/zfs/vdev_file.c
```diff
@@ -405,17 +466,23 @@ vdev_file_io_start(zio_t *zio)
...
 	LARGE_INTEGER offset;
-	void *b_data = NULL;
 
 	offset.QuadPart = zio->io_offset + vd->vdev_win_offset;
 
-	KeInitializeEvent(&Event, NotificationEvent, FALSE);
+	/* Preallocate space for IoWorkItem, required for vdev_file_io_start_done callback */
+	vf_callback_t *vb = (vf_callback_t *)kmem_alloc(sizeof(vf_callback_t) + IoSizeofWorkItem(), KM_SLEEP);
+
+	if (!vb) {
```
lundman, line 479: ditto

ZFSin/zfs/module/zfs/vdev_disk.c
aviksildc: sure, will do.

ZFSin/zfs/module/zfs/vdev_disk.c
aviksildc: ok, will remove.

aviksildc:
@lundman does this PR address https://github.com/openzfsonwindows/ZFSin/blame/master/README.md#L230-L232 ?

PR #270, opened 2020-06-24
Mark disks readonly/offline when ZFS has them in-use

lundman:
Which will stop the simple mistake of partitioning/formatting the wrong
disk. Note that users can still override it and corrupt the disk, but
similarly, users can dd to a raw disk in Unix.

We can decide if this is enough protection from the Users, or if we
should revisit this with more protection.


lundman:
@datacore-tlaurent

datacore-tlaurent:
I will try it.

datacore-tlaurent:
I'm closing this PR and creating a new one to reflect the final work done.

PR #273, opened 2020-06-25
Fixed kmem_bufctl_cache memory leak

arun-kv:
While unloading the zfs driver got  below error, this patch fixes this memory leak.

```
kmem_cache_destroy: 'kmem_bufctl_cache' (FFFFB0057DE02840) not empty
kmem_cache_destroy: 'kmem_slab_cache' (FFFFB0057DE02440) not emptySPL: Released 31 slabs
SPL: vmem_destroy('kmem_msb'): leaked 40960 bytes
SPL: vmem_destroy('kmem_metadata'): leaked 40960 bytes
```

### DIFF
--- a/ZFSin/spl/module/spl/spl-kmem.c
+++ b/ZFSin/spl/module/spl/spl-kmem.c
@@ -3921,6 +3921,30 @@ kmem_cache_set_move(kmem_cache_t *cp,
 }
 
 void
+kmem_qcache_destroy()
+{
+	kmem_cache_t *cp;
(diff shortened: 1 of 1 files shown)


imtiazdc:
@lundman There are more leaks than this change addresses. We are working on identifying and plugging them. In the meantime, please see if this change makes sense and please approve if it does.

lundman:
Ah we've known about this for a while over in osx, but never quite managed to locate where it was coming from, so this is very exciting.

PR #274, opened 2020-06-29
Enable WPP tracing in ZFSin project

vrajendra-datacore:
Enable WPP tracing to collect trace logs to an etl file

Below files are excluded from WPP tracing either because of format specifier "%.*S" not being supported by WPP or in case of arc.c, there is a macro arc_c defined in arc.c which does not allow the expansion of the function (dprintf or TraceEvent) in the desired way.

arc.c
dsl_proc.c
zfs_acl.c
zfs_ioctl.c
zfs_vfsops.c
zfs_vnops.c
zfs_vnops_windows.c
zfs_vnops_windows_lib.c
zfs_vnops_windows_mount.c
zfs_windows_zvol.c
zfs_znode.c

By default, the project settings of Release build has WPP enabled while that for Debug, it is disabled.
WPP can be enabled in Debug build as well if required.
To enable Wpp, Select Yes under 'Run Wpp Tracing' in Project properties; Define both RUN_WPP and WPPFILE="%(FileName).tmh" in C/C++ preprocessor

### DIFF
--- a/ZFSin/CMakeLists.txt
+++ b/ZFSin/CMakeLists.txt
@@ -109,6 +109,7 @@ add_subdirectory(zfs)
 wdk_add_driver(ZFSin
   debug.c
   driver.c
+  Wpp.c
 )
 
 target_link_libraries(ZFSin PRIVATE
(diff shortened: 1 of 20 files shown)


lundman:
Thanks, I'll look over this in the morning.

BTW, if you want to change a PR, you tend just to force push it again, and it updates. You don't need to create a new PR. You can add more commit, or squash them into one, then force push your branch and the PR updates.

No reason to change anything in this PR, but for the next time...

Issue #272, opened 2020-06-25
Honoring synchronous writes on ZVOL


aviksildc:
If a ZVOL is set with sync=standard (the default) and a write-through IO is sent to `zvol_write()`, it looks like it does not honor the write-through flag:
https://github.com/openzfsonwindows/ZFSin/blob/master/ZFSin/zfs/module/zfs/zvol.c#L2367

Ideally it should do something similar to its Linux counterpart: https://github.com/openzfs/zfs/blob/master/module/os/linux/zfs/zvol_os.c#L130-L131

The FUA bit can be checked from `pSrb->Cdb->CDBxx.ForceUnitAccess` as also explained here: https://disruptivesql.wordpress.com/2012/05/08/sata-and-write-through/

On the similar line, should `vdev_disk_io_start()` also honor the `ZIO_PRIORITY_SYNC_WRITE` priority while performing the writes? I see that a flag is set accordingly but eventually it is not carried forward:

https://github.com/openzfsonwindows/ZFSin/blob/master/ZFSin/zfs/module/zfs/vdev_disk.c#L575-L578

However in its Linux counterpart, this SYNC_WRITE priority is never checked.

lundman:
The ZVOL should be honoured, I just didn't know what the request would look like, so thanks for sharing the example.

aviksildc:
I think FUA is set at each write request when the device is opened with `FILE_FLAG_WRITE_THROUGH` set on `[Zw]CreateFile()`.

PR #280, opened 2020-07-10
Honouring ForceUnitAccess

lundman:
The illumos code uses WCE (Write Cache Enabled) in the Open call, to
set the style for future read/writes.

Windows appears to send ForceUnitAccess optionally with each write
request.

Retire the illumos setup as there appears not to be a Windows equivalent
flag during Open call. WCE is also retired in Linux.

Addressing #272


lundman:
flags to zvol_read() is just for symmetry.

aviksildc:
LGTM +1

PR #290, opened 2020-08-10
Trigger low_mem_event during zvol delete

arun-kv:
in zfs destroy, trigger low_mem_event, so that zfs could release all the unused memory to the system. Currently zfs ls holding the memory even when there is no zvol.


lundman:
I think you'd have to add this trigger a few more places that might release large parts of memory, rollback, clone, promote...  so it would probably be better if it was part of the periodic reap, as it is in illumos, so it works the same.

arun-kv:
Sure, i will check illumos code. 

First i thought of setting **lowmem = TRUE** in **spl_free_thread** if the system memory usage  and zfs **above_min_free_bytes** usage are above 80%. But i could not find any windows kernel api to get the system memory usage. That's the reason i just added only in zfs_destroy.

PR #295, opened 2020-08-20
Make all zdb commands working in ZFSin

aviksildc:
This patch makes all zdb commands working in ZFSin. This zdb util
leverages zpool cache file for running the commands.

This patch proposes to store the zpool cache file in
"C:\Windows\System32\drivers\zpool.cache". It allows running the
zdb commands without exporting the zpools.

The patch introduces zfs_file_* interfaces and deprecates kobj
interfaces. This is inspired by ZoL commit da92d5cb:
https://github.com/openzfs/zfs/commit/da92d5cbb38cea3a860b8a6bb8ee21f9129e7d7c
Although it does not remove vnode operations at the moment.

TODO: Fix the hang in kernel_fini caused by arc_flush.

Reviewed-by: Avik Sil <avik.sil@datacore.com>


lundman:
Ooo very nice, gimme a bit to read through it

ZFSin/zfs/cmd/zdb/zdb.c
```diff
@@ -2462,43 +2464,56 @@ dump_config(spa_t *spa)
 static void
 dump_cachefile(const char *cachefile)
 {
-	int fd;
-	struct stat statbuf;
 	char *buf;
 	nvlist_t *config;
-
-	if ((fd = open(cachefile, O_RDONLY)) < 0) {
```
lundman, line 2470: We might need to #ifdef WIN32 here -

ZFSin/zfs/cmd/zdb/zdb.c
```diff
@@ -2462,43 +2464,56 @@ dump_config(spa_t *spa)
...
+	err = zfs_file_getattr(&hFile, &attr);
+	if (err) {
 		(void) printf("failed to stat '%s': %s\n", cachefile,
-		    strerror(errno));
+	    strerror(errno));
 		exit(1);
 	}
 
-	if ((buf = malloc(statbuf.st_size)) == NULL) {
+	if ((buf = (char*)umem_alloc(attr.zfa_size, UMEM_NOFAIL)) == NULL) {
```
lundman, line 2496: If needed, should we wrap our own version of malloc() to call umem_alloc? Is umem_alloc() required here?

lundman: Another option is to to have the whole function as separate windows version.

ZFSin/zfs/include/sys/fs/zfs.h
```diff
@@ -805,7 +805,7 @@ typedef struct zpool_load_policy {
  * The location of the pool configuration repository, shared between kernel and
  * userland.
  */
-#define	ZPOOL_CACHE		"/etc/zfs/zpool.cache"
+#define	ZPOOL_CACHE		"\\SystemRoot\\System32\\drivers\\zpool.cache"
```
lundman, line 808: Should have #ifdef WIN32  # else  around it.

ZFSin/zfs/module/zfs/vdev_disk.c
```diff
@@ -162,9 +162,10 @@ vdev_disk_open(vdev_t *vd, uint64_t *psize, uint64_t *max_psize,
 
 	// Use vd->vdev_physpath first, if set, otherwise
 	// usual vd->vdev_path
-	vdev_path = vd->vdev_path;
 	if (vd->vdev_physpath)
-		vdev_path = vd->vdev_physpath;
+		vdev_path = spa_strdup(vd->vdev_physpath);
+	else
+		vdev_path = spa_strdup(vd->vdev_path);
```
lundman, line 168: vdev_path will leak in two error cases.

ZFSin/zfs/module/zfs/vdev_file.c
```diff
@@ -261,21 +278,34 @@ vdev_file_open(vdev_t* vd, uint64_t* psize, uint64_t* max_psize,
...
+		while (end && end[0] == '#') end++;
+		ddi_strtoull(end, &end, 10, &vd->vdev_win_length);
+		while (end && end[0] == '#') end++;
+
+		FileName = end;
+	} else
+		FileName = vd->vdev_path;
+
+	err = zfs_file_open(FileName, vdev_file_open_mode(spa_mode(vd->vdev_spa)), 0, &fp);
+	if (err) {
```
lundman, line 302: OK, adding the partition/len hack to vdev_file so zdb can read disks right?

ZFSin/zfs/include/sys/arc.h
```diff
@@ -265,6 +265,7 @@ void l2arc_init(void);
 void l2arc_fini(void);
 void l2arc_start(void);
 void l2arc_stop(void);
+void l2arc_stop2(void);
```
lundman, line 268: Left in by accident?

ZFSin/zfs/cmd/zdb/zdb.c
datacore-yash: will add

ZFSin/zfs/cmd/zdb/zdb.c
datacore-yash: Most of the userland programs use `umem_alloc` and it is a common code https://github.com/openzfsonwindows/ZFSin/blob/master/ZFSin/zfs/lib/libspl/include/umem.h#L84-L94
that's why we have used `umem_alloc`

ZFSin/zfs/include/sys/fs/zfs.h
datacore-yash: Will add

ZFSin/zfs/module/zfs/vdev_disk.c
datacore-yash: Will fix

ZFSin/zfs/module/zfs/vdev_file.c
datacore-yash: We have checked this partition/len hack in the above `if` block and tested it also.

ZFSin/zfs/include/sys/arc.h
datacore-yash: Will remove it.

ZFSin/zfs/include/sys/fs/zfs.h
```diff
@@ -805,8 +805,11 @@ typedef struct zpool_load_policy {
  * The location of the pool configuration repository, shared between kernel and
  * userland.
  */
+#if defined(WIN32) || defined(_KERNEL)
```
lundman, line 808: If win32 AND kernel maybe?

lundman (changes requested):
But for the one conditional it looks good!

ZFSin/zfs/include/sys/fs/zfs.h
datacore-yash: ZFSin driver (_KERNEL) writes to the cache file in kernel mode. And zdb reads the cache file in user mode (WIN32). That's why I've made it ||

lundman: But that means UNIX compile will also use use C:/Windows/ ...

datacore-yash: should it be like this `#if defined(WIN32) || (defined(_WIN32) && defined(_KERNEL))` ?

lundman (approved):

PR #301, opened 2020-09-02
Creates Logman session using zfsinstaller trace command

vrajendra-datacore:


vrajendra-datacore:
This fix provides a new command 'trace' for zfsinstaller.exe to create a logman (auto)session.

Once driver is installed, user can issue the command "zfsinstaller trace -f Flags -l Levels -s SizeOfETLInMB -p AbsolutePathOfETL" to create a logman session.
After reboot, the logs get collected in the path (AbsolutePathOfETL).
Also, handled are some scenarios
i)  If the provided Etl (path) exists already (We rename the existing etl and go ahead with provided etl file name)
ii) We try not to create logman session when one already exists. 

The etl file size is restricted to 'SizeOfETLInMB'. When it exceeds 'SizeOfETLInMB', older logs gets removed so as to maintain the etl size to 'SizeOfETLInMB'

The session gets deleted when the driver is uninstalled.

Utilized getopt function from libspl library for parsing the arguments; Have also changed the order of build process for zfsinstaller project

Issue #299, opened 2020-09-02
Are arc tuneables read during runtime?


chid:
I have been running ZFSin for a bit over two weeks and I've noticed that it consumes a huge chunk of my memory (20GB) without any way for me to reduce the amount. I have tried to set the zfs_arc_max to a much lower figure (500MB) and it appears that it's not honored. I have now adjusted it using the registry settings.

I expect that the answer would be to export and then reimport though I have had bluescreens the last few times I've tried to do this.

I noticed there was a similar issue occurring with ZFS https://github.com/openzfs/zfs/pull/8463 though that code block doesn't appear to exist in ZFSin.

lundman:
If you want to change things dynamically (without rebooting), you can use the kstat tool;

https://openzfsonosx.org/wiki/Windows_kstat

chid:
I have tried using the command `kstat.exe -w zfs:0:tunable:zfs_arc_max=5368709` and it didn't seem to have an impact. This is on version `OpenZFSOnWindows-release-20200814.exe`

lundman:
the zfs_arc_max is the write variable - then you read the new value from the other arc_max variable - i dont know why they split them. There is also a variable to simulate memory pressure - to make arc deflate

chid:
I assume you mean the other values are c, c_min, c_max

```
(base) PS C:\Windows\system32> kstat.exe -w zfs:0:tunable:zfs_arc_max=5368709
zfs:0:tunable:zfs_arc_max: 0 -> 5368709
(base) PS C:\Windows\system32> kstat.exe zfs:0:arcstats:c
module: zfs                             instance: 0
name:   arcstats                        class:    misc
        c                               16060096512

(base) PS C:\Windows\system32> kstat.exe zfs:0:arcstats:c_max
module: zfs                             instance: 0
name:   arcstats                        class:    misc
        c_max                           16060096512

(base) PS C:\Windows\system32> kstat.exe zfs:0:arcstats:c_min
module: zfs                             instance: 0
name:   arcstats                        class:    misc
        c_min                           2007512064

(base) PS C:\Windows\system32> kstat.exe -w zfs:0:tunable:zfs_arc_max=2007512064
zfs:0:tunable:zfs_arc_max: 5368709 -> 2007512064
(base) PS C:\Windows\system32> kstat.exe zfs:0:arcstats:c_min
module: zfs                             instance: 0
name:   arcstats                        class:    misc
        c_min                           2007512064

[...]

PR #310, opened 2020-10-12
Fix arc kstat updates

datacore-raj:
This patch partially reverts commit 88d37b7a1ae78ae36bfac16e7602885f54b266cc
The issue was that the os_mem_alloc was keeping on increasing when IO
workload was running, not honouring zfs_arc_max and zfs_arc_meta_limit.

### DIFF
--- a/ZFSin/zfs/module/zfs/zfs_kstat_windows.c
+++ b/ZFSin/zfs/module/zfs/zfs_kstat_windows.c
@@ -216,6 +216,10 @@ static int osx_kstat_update(kstat_t *ksp, int rw)
 		zfs_vnop_skip_unlinked_drain = ks->win32_skip_unlinked_drain.value.ui64;
 		zfs_vfs_sync_paranoia = ks->win32_use_system_sync.value.ui64;
 
+		/* ARC */
+		arc_kstat_update(ksp, rw);
+		arc_kstat_update_osx(ksp, rw);
+
 		/* L2ARC */
 		l2arc_write_max = ks->l2arc_write_max.value.ui64;
 		l2arc_write_boost = ks->l2arc_write_boost.value.ui64;
@@ -410,6 +414,10 @@ static int osx_kstat_update(kstat_t *ksp, int rw)
 		ks->win32_force_formd_normalized.value.ui64 = zfs_vnop_force_formd_normalized_output;
 		ks->win32_skip_unlinked_drain.value.ui64    = zfs_vnop_skip_unlinked_drain;
 		ks->win32_use_system_sync.value.ui64 = zfs_vfs_sync_paranoia;
+
+		/* ARC */
+		arc_kstat_update(ksp, rw);
+		arc_kstat_update_osx(ksp, rw);
 
 		/* L2ARC */
 		ks->l2arc_write_max.value.ui64               = l2arc_write_max;


chid:
Looks like this will fix #299

lundman:
Ah yes - the logic for setting `zfs_arc_max` is no longer run, but the reason it was taken out, is that `ksp` when called is for `osx_kstat_update()` is a `osx_kstat_t osx_kstat`, then we pass that to `arc_update()` as if it is `kstat_t			*arc_ksp`. Quite naughty. A quick fix would be to change to:
```
	arc_kstat_update(arc_ksp, rw);
	arc_kstat_update_osx(ksp, rw);
```

datacore-raj:
> Ah yes - the logic for setting `zfs_arc_max` is no longer run, but the reason it was taken out, is that `ksp` when called is for `osx_kstat_update()` is a `osx_kstat_t osx_kstat`, then we pass that to `arc_update()` as if it is `kstat_t *arc_ksp`. Quite naughty. A quick fix would be to change to:
> 
> ```
> 	arc_kstat_update(arc_ksp, rw);
> 	arc_kstat_update_osx(ksp, rw);
> ```

datacore-raj:
With above changes (using `arc_ksp` instead of `ksp` in `arc_kstat_update()` ) not able to install ZFSin as `arc_ksp` initialize after this.
It crashes with stack trace:-

ZFSin!arc_kstat_update+0xd [C:\ZFS-vt-code\ZFSin\ZFSin\zfs\module\zfs\arc.c @ 7329] 
ZFSin!osx_kstat_update+0x5a7 [C:\ZFS-vt-code\ZFSin\ZFSin\zfs\module\zfs\zfs_kstat_windows.c @ 430] 
ZFSin!kstat_osx_init+0xa0 [C:\ZFS-vt-code\ZFSin\ZFSin\zfs\module\zfs\zfs_kstat_windows.c @ 625] 
ZFSin!DriverEntry+0x2e [C:\ZFS-vt-code\ZFSin\ZFSin\driver.c @ 75] 
ZFSin!FxDriverEntryWorker+0xa9 [d:\5359\minkernel\wdf\framework\kmdf\src\dynamic\stub\stub.cpp @ 287] 
nt!IopLoadDriver+0x4bd
nt!IopLoadUnloadDriver+0x4a
nt!ExpWorkerThread+0x16a
nt!PspSystemThreadStartup+0x55
nt!KiStartSystemThread+0x1c

lundman:
Yeah, we would need to find a solution. Although, the zfs_arc_max and friends are updated in arc_kstat_update_osx(), perhaps the answer is to just bring that one back.

imtiazdc:
@lundman Is something wrong with this PR? Or the build? Please let me know how we can fix - "All checks have failed".

If nothing is wrong with the PR, can this change be approved?

lundman:
Ah yes, it is wrong and I was expecting an update. But it might be easier if I update it? We do need to set the values, this is true, but the old code passes the wrong struct to arc update - its pure luck that struct is big enough to fit all the writes.

imtiazdc:
> But it might be easier if I update it?

@lundman It would be wonderful if you could provide the update. We will be happy to test it and update you with our findings. We don't have enough awareness about that part of code to fix it ourselves in the near term.

lundman:
It comes down to:
`static int osx_kstat_update(kstat_t *ksp, int rw)`
which expects that ksp to be:
```
typedef struct osx_kstat {
	kstat_named_t spa_version;
	kstat_named_t zpl_version;
```
and that update function does indeed update those members.
I believe this is correct (hurrah), but then we call

```
		arc_kstat_update(ksp, rw);
		arc_kstat_update_osx(ksp, rw);
```
using the same `ksp`.

In this case:
`arc_kstat_update(kstat_t *ksp, int rw)`
expects the struct
```
typedef struct arc_stats {
	kstat_named_t arcstat_hits;
	kstat_named_t arcstat_misses;
	kstat_named_t arcstat_demand_data_hits;
```
So when `arc_kstat_update()` updates `arcstat_hits` it is actually overwriting `spa_version`! Yikes.

Later on, kstat calls `arc_kstat_update()` with the correct `ksp` - so values are put back to real ones. You'd have to be pretty lucky with timing to "see" it running kstat, since the calls are right after each other. But really, we are writing ints, and strings, over struct members that are neither, then writing the correct ones back.

imtiazdc:
> It comes down to:
> `static int osx_kstat_update(kstat_t *ksp, int rw)`
> which expects that ksp to be:
> 
> ```
> typedef struct osx_kstat {
> 	kstat_named_t spa_version;
> 	kstat_named_t zpl_version;
> ```
> 
> and that update function does indeed update those members.
> I believe this is correct (hurrah), but then we call
> 
> ```
> 		arc_kstat_update(ksp, rw);
> 		arc_kstat_update_osx(ksp, rw);
> ```
> 
> using the same `ksp`.
> 
> In this case:
> `arc_kstat_update(kstat_t *ksp, int rw)`
> expects the struct
> 
> ```
> typedef struct arc_stats {
> 	kstat_named_t arcstat_hits;
> 	kstat_named_t arcstat_misses;
> 	kstat_named_t arcstat_demand_data_hits;
> ```
> 
> So when `arc_kstat_update()` updates `arcstat_hits` it is actually overwriting `spa_version`! Yikes.
> 
> Later on, kstat calls `arc_kstat_update()` with the correct `ksp` - so values are put back to real ones. You'd have to be pretty lucky with timing to "see" it running kstat, since the calls are right after each other. But really, we are writing ints, and strings, over struct members that are neither, then writing the correct ones back.

Hmm, interesting. Will you be able to submit the fix sometime soon? Say in a week? If not, I will have someone take a look at this thread and raise a PR. Please do let me know.

lundman:
Yeah, I'll push something tomorrow

imtiazdc:
> Yeah, I'll push something tomorrow

@lundman Are there more changes in the pipeline? With [this](https://github.com/openzfsonwindows/ZFSin/commit/9bf81599f0f02cc9f4226bf2a8f4651af2357940) commit alone, we still don't see the arc_max limit being honored.

Total RAM: 128 GB
ZFSin memory usage : 63.4 GB
zfs_arc_max : 15 GB

lundman:
Through other commits I think we have corrected this. Please re-open if issues remain.

imtiazdc:
> Through other commits I think we have corrected this. Please re-open if issues remain.

Thanks, @lundman. Yes, your latest checkins do honor the arc_max limit. The arc_max limit does get violated for a brief period but then the arc usage falls below the arc_max within a few seconds.

PR #314, opened 2020-11-06
fix for zpool import

datacore-yash:
release dn_struct_rwlock in dmu_zfetch codepath


lundman:
You can just force-push the correct branch to the PR to update it...

Issue #312, opened 2020-10-19
Import hangs with build 1.0.216


adamdmoss:
My pool import hangs with https://ci.appveyor.com/project/lundman/zfsin/builds/35813886 - zpool doesn't return and the datasets aren't mounted/available.
1.0.213 / https://ci.appveyor.com/project/lundman/zfsin/builds/35700991 seems good.

adamdmoss:
Sorry for lack of details, I wanted to get back to a working system without spending too much time diagnosing.
FWIW I switched between 1.0.213 and 1.0.216 a few times to verify.

lundman:
Hmm so a recent break? Thanks for reporting it

PR #315, opened 2020-11-06
release dn_struct_rwlock in dmu_zfetch codepath

datacore-yash:
This patch fixes https://github.com/openzfsonwindows/ZFSin/issues/312


lundman (approved):
Upstream have these, I wonder how I managed to lose them..  This looks good, cheers.

PR #324, opened 2021-01-12
Fixed a race condition in 'spl_cv_wait' that could potentially cause a thread to miss a broadcast event

arun-kv:
Moved mutex_enter above waiters_count == 1 check in spl_cv_wait, to avoid the possibility of current thread wrongly detecting that is the only thread waiting for the event while there could be another  thread that acquired the mutex and is about to increment the  'waiters_count' a few lines above.

### DIFF
--- a/ZFSin/spl/include/osx/condvar.h
+++ b/ZFSin/spl/include/osx/condvar.h
@@ -19,7 +19,7 @@ enum {
 struct cv {
 	KEVENT kevent[CV_MAX_EVENTS]; // signal event, broadcast event
 	KSPIN_LOCK waiters_count_lock;
-	uint32_t waiters_count;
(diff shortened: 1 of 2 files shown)


imtiazdc:
@lundman For some background, we ran into an issue where the dump seems to indicate there is at least one thread waiting endlessly choking other threads depending on it. 

I spent some thought with @arun-kv on this change and this PR looks good to me. Is there any specific test you want us to run as this change is at the core of synchronization in ZFSin?

lundman:
That looks quite reasonable. I suspect there is further issue, I have had `mutex_exit()` panic due to `not held`. But it happens so infrequently it has been hard to track.

lundman (approved):

PR #326, opened 2021-01-19
fix zdb -R for raw flag

datacore-skumar:
Write was returing wrong size due to some non printable characters
in buffer So, I replaced Write with putchar.

### DIFF
--- a/ZFSin/zfs/cmd/zdb/zdb.c
+++ b/ZFSin/zfs/cmd/zdb/zdb.c
@@ -5061,7 +5061,13 @@ zdb_dump_block_raw(void *buf, uint64_t size, int flags)
 {
 	if (flags & ZDB_FLAG_BSWAP)
 		byteswap_uint64_array(buf, size);
+#ifdef _WIN32
+	HANDLE hStdout;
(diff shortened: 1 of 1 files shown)


lundman:
Is this the sort of place that O_BINARY would affect the outcome?

datacore-skumar:
The issue was with wrong file handle to `WriteFile()`. It's working fine after fixing the file handle.

lundman:
Ah yep, the sort of thing "wosix.h" is supposed to hide, but it probably isn't used for zdb.  This will do fine in ZFSin, but let's take the wosix approach in the new port..

lundman (approved):

lundman:
What is the status here?

imtiazdc:
> What is the status here?

@lundman This could be merged in ZFSin. We tested this internally and found it working.

lundman:
Ok, posix.c and wosix.h is supposed to wrap the `write(_fileno(stdout),` parts, with fileno going to HANDLE, and write using WriteFile(HANDLE, ...  (although, its supposed to be fileno, not _fileno, since that will call Windows fileno and return a filedes.

GAMBOT8:
Folks, I require assistance. Anyone may assist me?

PR #327, opened 2021-01-20
Periodically calling 'vmem_update' function to dynamically update vmem->vm_hash_table size

arun-kv:
Currently 'vmp->vm_hash_table' size is 16, due to which the lookup time is too high when the number of objects are more. We have seen ZFSin uninstallation taking more than (~30 min) when the arc usage is ~40GB. 
During the analysis we saw the while loop inside 'vmem_hash_delete'  taking most of the time.
prev_vspp = VMEM_HASH(vmp, addr);
while ((vsp = *prev_vspp) != NULL) {
                if (vsp->vs_start == addr) {
                        *prev_vspp = vsp->vs_knext;
                        break;
                }
                vmp->vm_kstat.vk_lookup.value.ui64++;
                prev_vspp = &vsp->vs_knext;
        }

So enabled 'vmem_update' function to periodically call 'vmem_hash_rescale' to update the hash_table size every 60 sec. After the fix, vm_hash_table size is much bigger and the lookup time has been reduced considerably, also the uninstallation process is very fast.

### DIFF
--- a/ZFSin/spl/module/spl/spl-kmem.c
+++ b/ZFSin/spl/module/spl/spl-kmem.c
@@ -130,6 +130,8 @@ static const char *KMEM_VA_PREFIX = "kmem_va";
 static const char *KMEM_MAGAZINE_PREFIX = "kmem_magazine_";
 
 static char kext_version[64] = SPL_META_VERSION "-" SPL_META_RELEASE SPL_DEBUG_STR;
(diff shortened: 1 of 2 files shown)


lundman (approved):

imtiazdc:
@lundman Could you merge this change please?

lundman:
Ah so sorry for the delay

PR #340, opened 2021-07-20
 Fixed ZFSin going to an infinate loop when the underlying disk returns error

arun-kv:
Fixed ZFSin going to an infinate loop when the underlying disk returns error, and causing data loss.
  Also fixed 'zpool clear' command not working on ZFSin.

### DIFF
--- a/ZFSin/zfs/module/zfs/zio.c
+++ b/ZFSin/zfs/module/zfs/zio.c
@@ -2130,7 +2130,7 @@ zio_suspend(spa_t *spa, zio_t *zio, zio_suspend_reason_t reason)
 		    "failure and the failure mode property for this pool "
 		    "is set to panic.", spa_name(spa));
 
-	cmn_err(CE_WARN, "Pool '%s' has encountered an uncorrectable I/O "
+	dprintf("Pool '%s' has encountered an uncorrectable I/O "
 	    "failure and has been suspended.\n", spa_name(spa));
 
 	zfs_ereport_post(FM_EREPORT_ZFS_IO_FAILURE, spa, NULL,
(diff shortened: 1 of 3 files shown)


ZFSin/zfs/module/zfs/zio.c
```diff
@@ -3605,7 +3605,7 @@ zio_vdev_io_done(zio_t *zio)
 
 	ops->vdev_op_io_done(zio);
 
-	if (unexpected_error && zio->io_waiter != NULL)
+	if (unexpected_error)
```
lundman, line 3608: Do we know why we have to make this change to the upstream code?

arun-kv: When there is an EIO error coming from the disk ZFSin is currently not calling vdev_probe, which is actually responsible for putting the zpool into suspended state(by calling zio_suspend). Now in ZFSin we are repeatedly retrying to sync the transaction. The right behavior is to put the zpool into suspended state and the after fixing the disk error run 'zpool clear' command to remove suspend state.  So with **zio->io_waiter != NULL** check vdev_probe no getting called even though there is an error(I actually found this fix from openzfs).

Without this fix we will have a data miscompare issue, as my application is getting write success message even when zfs is getting EIO from the underlying disk(with sync=always and application is using NO_BUFFERING flag). With this fix we still have a miscompare, but its only for the first write after the disk error, before this fix all the writes after the disk error was wrongly acknowledged. The first write miscompare after disk error  is present in openzfs also,  I have also filed a bug in openzfs  **https://github.com/openzfs/zfs/issues/12391**.

lundman: Ah perfect, thanks.

